Issue #1, opened 2014-06-06
Markdown support


methane:
Recent days, many python project have README in Markdown format.

AraHaan:
I for one support this as it is a pain having to convert tables that can be in the readme to restructured text that can very easily break the tables. This would then brevent such breakage of them.

JulienPalard:
Having just pushed:

![Markdown vs RestructuredText pie](https://github.com/JulienPalard/pystyle-data/raw/master/plots/readme.png)

I searched and found this issue. It does not look impossible to classify markdown vs restructurextext documents, having a module to tell them appart may be a good start that may be upgraded later (like getting the information from the setup instead of guessing?), what do you think? Should I try to do it?

JulienPalard:
I tried if it was possible to tell appart md or rst with the content only, looks not that hard:

https://github.com/JulienPalard/mdorrst

It's slightly biaised towards rst, which is not bad for this usage.

It may be used as a temporary solution for `setup.py` files not explicitly giving the format, what do you think?

ewdurbin:
@JulienPalard love the idea behind `mdorrst`, just as a headsup: for the time being if this is to be used as a dependency for `readme_renderer`, it will require python 2.7 support until pypa/pypi-legacy dies.

AraHaan:
Which is probably when 2.7 reaches EOL (if not extended yet again because of linux distributes requireing it to work properly).

AraHaan:
Another thing I want in the markdown support. The good old fashoned checkboxes like so: ``- [x]`` or ``- [ ]`` They can be so useful on checking when features of things are added (or partially added).

JulienPalard:
@ewdurbin Support for Python 2.7 done in mdorrst 0.3.0:
```
python2.7(tmp) mdk@lighthaven$ python2.7 -m pip install mdorrst
Collecting mdorrst
  Downloading mdorrst-0.3.0-py2.py3-none-any.whl
Installing collected packages: mdorrst
Successfully installed mdorrst-0.3.0
(tmp) mdk@lighthaven$ python2.7
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import mdorrst
>>> mdorrst.from_text("[hey](http://example.com)")
'md'
>>> mdorrst.from_text("`hey <http://example.com>`__")
'rst'
>>> 
```

[...]

PR #3, opened 2014-12-09
Add Markdown support

msabramo:
Fixes: GH-1

```
$ tox -e py27,py33,py34
GLOB sdist-make: /Users/marca/dev/git-repos/readme/setup.py
py27 inst-nodeps: /Users/marca/dev/git-repos/readme/.tox/dist/readme-0.3.0.zip
py27 runtests: PYTHONHASHSEED='2116262944'
py27 runtests: commands[0] | py.test tests/
============================================================================= test session starts ==============================================================================
platform darwin -- Python 2.7.6 -- py-1.4.26 -- pytest-2.6.4
plugins: capturelog, cov
collected 7 items

tests/test_markdown.py .......

=========================================================================== 7 passed in 0.18 seconds ===========================================================================
py33 inst-nodeps: /Users/marca/dev/git-repos/readme/.tox/dist/readme-0.3.0.zip
py33 runtests: PYTHONHASHSEED='2116262944'
py33 runtests: commands[0] | py.test tests/
============================================================================= test session starts ==============================================================================
platform darwin -- Python 3.3.3 -- py-1.4.26 -- pytest-2.6.4
plugins: capturelog, cov
collected 7 items

tests/test_markdown.py .......

=========================================================================== 7 passed in 0.22 seconds ===========================================================================
py34 inst-nodeps: /Users/marca/dev/git-repos/readme/.tox/dist/readme-0.3.0.zip
py34 runtests: PYTHONHASHSEED='2116262944'
py34 runtests: commands[0] | py.test tests/
============================================================================= test session starts ==============================================================================
platform darwin -- Python 3.4.0 -- py-1.4.26 -- pytest-2.6.4
plugins: capturelog, cov
collected 7 items

tests/test_markdown.py .......

=========================================================================== 7 passed in 0.20 seconds ===========================================================================
___________________________________________________________________________________ summary ____________________________________________________________________________________
  py27: commands succeeded
  py33: commands succeeded
  py34: commands succeeded
  congratulations :)
```

Passing Travis CI build: https://travis-ci.org/msabramo/readme/builds/43547312

Cc: @dstufft, @r1chardj0n3s, @hickford, @methane


msabramo:
@dstufft: If you have comments, I can update this.

dstufft:
I don't think it makes sense to merge any Markdown support until we have an answer for how PyPI (or any other tool) is supposed to know what format it's supposed to use.

tests/test_markdown.py
```diff
@@ -0,0 +1,83 @@
...
+
+    out, rendered = render(md_markup)
+
+    assert rendered == expected_rendered
+    assert out == expected_html.rstrip()
+
+
+def read(fn):
+    path = os.path.join(os.path.dirname(__file__), 'fixtures', 'markdown', fn)
+    with codecs.open(path, encoding='utf-8') as f:
```
berkerpeksag: Use `io.open` instead. `codecs.open` will be deprecated: http://bugs.python.org/issue8796

bukzor:
I don't think we should block this merge on "how PyPI (or any other tool) is supposed to know what format it's supposed to use"; we can merge this without regressing on that front, and there are a few obvious ways to improve it later.

tests/test_markdown.py
msabramo: Updated. Thanks!

msabramo:
@bukzor: I agree. I think that the rendering of Markdown vs. the decision of what markup language to render are orthogonal issues. 

It's possible that pypa/README could attempt to guess the format from a blob of text using some heuristics, but I think that the more robust solution is to have something at a higher level that guesses the format based on filename extension or even better, a new piece of distribution metadata -- e.g.: `long_description_content_type='text/markdown'`. 

I don't think figuring out the mechanism for choosing the renderer needs to block making progress on making Markdown rendering work nicely (which I think largely means having parity with GitHub-flavored Markdown).

sudarkoff:
I completely agree with @msabramo and @bukzor

dstufft:
What's the point of being able to render Markdown without being able to determine if we *should* render as Markdown? Without some way to determine what we should use this code is just going to sit there doing nothing.

msabramo:
Well yeah, I get that the Markdown stuff won't be used until we figure out how. But I think we should try to work on both. 

One could make the reverse argument -- What's the point of having a mechanism for identifying different content types and choosing between them when there is only one markup format?
I could argue that we're going through that now. There was some discussion a while back about how to specify content types and then it kind of stalled (as things in open source often do). But I wonder -- if we had the markdown rendering, that might help motivate the discussion. 

I don't like having dead code but my bigger concern would be that if we couple two efforts together then each one stalls the other and increases the chances of a deadlock.

dstufft:
We already have two rendering formats, plaintext and rst. The selection mechanism is implicit and fallback based. This won't work if we add in a third rendering format but anything that works for 3 formats will work for 2 formats.

A better mechanism to choose between them would mean that PyPI could reject uploads which specify rst and which fail at rst rendering (for instance). There's no dependency on Markdown for being able to be explicit about how you want the long description rendered. There is a dependency on being explicit about how to render the long description before we can actually render anything with Markdown.

bukzor:
@dstufft: You make some valid points, but I don't see it as an argument to block this merge, let this work bitrot. As @msabramo noted, once this is merged, everyone will be much happier to work on a content-type switching mechanism. It may not be the order your prefer, but it's certainly the order it's been worked on.

msabramo:
Well put, @bukzor!

I also think @dstufft made some valid points about how the content type switching could be useful without Markdown.

Which just confirms for me that these two things are independent of each other and could happen in either order. So why not just tackle them in the order that they naturally unfold so that we maintain forward momentum?

dstufft:
Content type switching is independent, additional markup language is dependent on content type switching.

msabramo:
@dstufft Well then I can look at adding support to Warehouse. Do you have a solution to https://github.com/pypa/warehouse/issues/343 ?

dstufft:
Yea I can get something later today, though the problem is less a code problem and more of a problem that someone needs to champion some method of detection on distutils-sig and get some consensus there.

bukzor:
This might be helpful?

http://aws.amazon.com/public-data-sets/

On Fri, Jan 30, 2015 at 12:31 PM, Donald Stufft <notifications@github.com>
wrote:

> Yea I can get something later today, though the problem is less a code
> problem and more of a problem that someone needs to champion some method of
> detection on distutils-sig and get some consensus there.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pypa/readme/pull/3#issuecomment-72265524>.
>



-- 
Buck Golemon
Cell: (408) 202-3901

msabramo:
@methane: I added the extensions you requested and also `markdown.extensions.codehilite`. Thanks for the suggestions!

msabramo:
~~~python
setup(
    long_description_filename='README.md',
    long_description_content_type='text/x-markdown; charset=UTF-8',
)
~~~

bukzor:
Manual support would be a great first step. More auto-magic can be added later. Some might argue it *should* come later.

A couple quick improvements, for afterward:

 1. content_type='text/x-markdown; charset=UTF-8' should be the content type guessed for a file ending with `.md`. I dimly recall some stdlib support for mimetype guessing.
 2. Extending the pypi auto-readme behavior to fall back to README.md, as laid out in https://bitbucket.org/pypa/pypi/pull-request/33

msabramo:
https://docs.python.org/2/library/mimetypes.html#mimetypes.guess_type

nicktimko:
Is this dead/not going to happen for organizational reasons (versus anything technical) or still being considered?

karimbahgat:
Same question: is this merge still happening, given that the tests were succesfull? I realize as stated that the other half of the problem is that distutils needs to also be changed to accept a long_description_content_type, so does anyone know if there has been any progress/issue/pull-request in the distutils repo? 
Finally, assuming the markdown rendering change and the distutil change were both accepted, how long would it realistically take before pypi would begin rendering markdown text (assuming the developer published their package with the most recent version of distutils)? I.e. how often are new official distutils versions released?

bukzor:
bump =/

msabramo:
I'm willing to update this according to comments, but only if there is a reasonable chance of it being merged.

nchammas:
> the problem is less a code problem and more of a problem that someone needs to champion some method of detection on distutils-sig and get some consensus there.

It's good that we have some code ready to go here, but it sounds like what the project needs is not someone to write code just yet, but rather someone to lead a discussion and build a consensus on how we're supposed to specify/detect the README format across the various packaging tools that deal with READMEs.

@msabramo / @bukzor / others interested in this issue: Have you considered leading a discussion on [distutils-sig](https://mail.python.org/mailman/listinfo/distutils-sig/) to build the consensus we need?

I know it's annoying to have to sign up for a mailing list just to discuss an issue, and building consensus on an approach to take is not an easy task, but that's what's needed here.

illume:
Pretty sure close to 100% of ('README.md', '*.md') packages are using markdown format.
Pretty sure.
Pretty sure this bikeshed issue does not need more discussion.
Pretty sure.
Pretty sure all the .md readmes on pypi and devpi are still broken.
Pretty sure.
Pretty sure this is a safe first start. Unless the powers that be don't want Markdown at all.
Pretty sure.

nicktimko:
Here's [a stab at it](https://mail.python.org/pipermail/distutils-sig/2016-May/028677.html). Not sure if there's some more official proposal method/format, but if the deciders just follow the mailing list then hopefully it'll capture their eyeballs.

nicktimko:
So I guess in my zeal to try to get something working ASAP and talking about how it should look, I set up something of a straw-man that got smashed apart. 

* (2) people opposed it on grounds that GitHub, et al. also support rST, so just use that anyways
* (1) person pointed out that "Markdown" has no standard (but maybe CommonMark, but then maybe not the semi-ubiquitous GFM).

Some other mentions of "just install pandoc + pypandoc". It seems like there's a rift between the GH issue tracker and the actual-official mailing lists.

karimbahgat:
That sounds typical. 

The first argument is really besides the point, since maybe the main reason people want markdown is its ease of use and simplicity, vs rst which can be a bit of a challenge to learn and restricted to the more technically adept. Just because some devout enthusiasts and skilled scientists know how to do something in a complex way doesnt mean its easy for everyone else. There are also large swaths of programmers who are good at and enjoy and contribute packages to python but arent interested or have the time to learn multiple different markup, or even if they knew would rather go for the easiest option. Thats the whole point of python in the first place. 

Which relates to the third argument, requiring nonpython installations like pandoc that also require some level of comfertability and understanding of commandline use and c compiling. Thats enough to scare away many ad hoc programmers, or just make their pypi documentation look shitty. 

As for the second argument, i think theres a pretty good chunk of core markdown that doesnt change between the various specifications, eg from the guy who first proposed created markdown. Providing support for those basics should be more than enough. 

I really like the suggestions youve put forth, they seem simple, doable, and wont make anything backwards incompatible. 

I would post this to the mailing lists, but as a case in point i dont know how to or have the time.

nicktimko:
I got an update on this from the PyPA core devs a while ago and haven't had time to relay to here, but there's no real opposition, but it needs to be treated holistically rather than a single thing. Here are [the steps advised by @ncoghlan on the mailing list](https://mail.python.org/pipermail/distutils-sig/2016-June/029078.html) to take in order to add new markup format(s):

* Update the metadata spec for the Python Packaging User Guide (PyPUG) ([hosted](https://packaging.python.org/en/latest/specifications/#package-distribution-metadata) & [source](https://github.com/pypa/python-packaging-user-guide/blob/master/source/specifications.rst)) to document the new field (e.g. `Description-Format: txt|rst|md|whatever`). This document has built on PEP-345 ([which is where `Description:` is documented](https://www.python.org/dev/peps/pep-0345/#description-optional)) and is the new de-facto spec.

* [`pypa/setuptools`](https://github.com/pypa/setuptools/tree/master/setuptools) to support setting it (i.e. adding an argument to `setup()`?)

* [`pypa/warehouse`](https://github.com/pypa/warehouse) to respect it for all defined rendering formats

* (potentially) [`pypa/pypi-legacy`](https://github.com/pypa/pypi-legacy) to support respecting it at least for reStructured text and Markdown (treating others as plain text)

     * (haven't checked, but I'm assuming that the above two call back to this `pypa/readme_renderer` package/repo?)

* PyPUG again to provide usage docs once the default tools support it

Many hoops, but it does seem to make sense because the whole packaging/distribution system is broken up across specifications, client-software, server-software (new and old).

msabramo:
Just fixed the merge conflict in this PR.

msabramo:
@nicktimko: Thank you so much for taking the lead here and finding out the steps that are necessary to make this happen! That is super helpful!

For the first item you listed, "Update the metadata spec for the Python Packaging User Guide (PyPUG)", I just submitted a [PR to the PyPUG](https://github.com/pypa/python-packaging-user-guide/pull/258) to propose a new `Description-Content-Type` metadata field: https://github.com/pypa/python-packaging-user-guide/pull/258

msabramo:
Hmm, why are there no Travis CI test results for this PR?

msabramo:
@dstufft: Are you able to trigger the Travis tests?

msabramo:
Now that we seem to be gravitating towards making CommonMark the preferred Markdown variant in https://github.com/pypa/python-packaging-user-guide/pull/258, I should probably test how well this renders CommonMark and possibly modify it to be more CommonMark-compliant (perhaps using https://pypi.python.org/pypi/CommonMark ?)

msabramo:
Checklist with steps that @nicktimko mentioned:

- [ ] Update the metadata spec for the Python Packaging User Guide (PyPUG) (hosted & source) to document the new field -- in progress -- https://github.com/pypa/python-packaging-user-guide/pull/258

- [ ] pypa/setuptools to support setting it -- in progress -- https://github.com/pypa/setuptools/pull/708

- [ ] pypa/warehouse to respect it for all defined rendering formats -- in progress -- this PR

- [ ] (potentially) pypa/pypi-legacy to support respecting it at least for reStructured text and Markdown (treating others as plain text) (haven't checked, but I'm assuming that the above two call back to this pypa/readme_renderer package/repo?)

- [ ] PyPUG again to provide usage docs once the default tools support it

AraHaan:
an option is to look at and see if rst and markdown both support some method of commenting stuff out. And then if they both have it we could use it for an advantage and look for comments containing either ``markdown`` or ``restructuredtext`` (case insensitive) at the top of the files/data sent to pypi.

msabramo:
Time has shown that I haven't been able to pick this up (in part because I've changed roles and Python isn't currently a big part of my new role). Anyone else want to pick this up?

AraHaan:
I got an even better idea on it, why not use the vim style syntax rendering.

like then you could tell it if it is ``md`` or ``rst``?

tests/test_markdown.py
```diff
@@ -0,0 +1,95 @@
+import io
```
AraHaan: Encoding at top of file? UTF8 etc?

tests/fixtures/markdown/code.md
```diff
@@ -0,0 +1,8 @@
+Here is some Python:
+
+```python
```
AraHaan: you can exchange \`\`\`python with \`\`\`py and it would render the same.

JulienPalard:
We now have a way from wharehouse filters to tell appart markdown and rst via https://pypi.python.org/pypi/mdorrst, does something else blocks this PR?

Lucas-C:
There seems to be conflicts now.
But please : finish this PR ! This is an awesome feature !

JulienPalard:
see https://github.com/pypa/readme_renderer/pull/51 also.

PR #13, opened 2014-12-18
Allow the `class` attr for all tags

msabramo:
because we want to use this module in PyPI and PyPI's current generated HTML has these `class` attributes.


msabramo:
PyPI is already using pypa/readme, so I think this can be closed, since it has conflicts anyway.

PR #14, opened 2014-12-19
Disable initial_header_level=2

msabramo:
PyPI needs this to be 1 (the default). So we either need to remove this or make it configurable.


msabramo:
Well, PyPI is using pypa/readme now, so if we're happy with the way stuff looks in PyPI now, then I guess this can be closed.

PR #15, opened 2014-12-19
Make clean and linkify configurable

msabramo:
`clean` and `linkify` are optional parameters to the `render` function. The caller can provider their own callable for these or they can set them to `None` to turn them off completely.

PyPI, in particular, needed the ability to turn `linkify` off, because it's adding extra `rel="nofollow"` attributes to links that it doesn't have (but maybe should when it switches to using pypa/readme?)

### DIFF
--- a/readme/clean.py
+++ b/readme/clean.py
@@ -43,30 +43,10 @@ ALLOWED_STYLES = []
 
 
 def clean(html):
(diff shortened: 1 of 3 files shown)


dstufft:
I don't believe that readme should have any configuration options. It should do one thing and only one thing.

PR #16, opened 2014-12-19
Disable smart quotes

msabramo:
because PyPI doesn't use them.

We can turn it on later perhaps.


dstufft:
I want to keep the smart quotes around.

PR #51, opened 2017-04-27
Markdown

JulienPalard:
On top of the work of @msabramo on https://github.com/pypa/readme_renderer/issues/1, I added my [mdorrst](https://github.com/JulienPalard/mdorrst) to properly dispatch to the right renderer.

This introduces `readme_renderer.any.render`, dispatching itself to `readme_renderer.rst.render`, `readme_renderer.markdown.render`, and `readme_renderer.txt.render`.

`any` may not be the best choice as it conflicts with a builtin, but only an "import *" or an "import readme_renderer.any as any" can override the builtin, so I prefer the verbosity of the name, is someone has better in mind just tell me. We may also "from .any import render" in `__init__.py` so other projects will simply use "readme_renderer.render(long_description)".

/!\ This renderer *in fact* modifies the rendering of some existing README:
Before, we were always trying to render long_descriptions as rst, even if it was plaintext or HTML or whatever, we were fallbacking only if rst rendering was failing.

Now, this branch simply use the txt renderer if the content is "obviously only text". It may be what we want, it may not, (and I'm sure it's we want for half of them, and not what we want for the other half...).


tests/fixtures/markdown/code.md
```diff
@@ -0,0 +1,8 @@
+Here is some Python:
+
+```python
```
AraHaan: change this to \`\`\`py. This is a shortcut for \`\`\`python.

JulienPalard: As it's a text fixture I really don't see the point of changing this. If I touch it, maybe I can add a new block with py so the test ensuer both?

AraHaan: Yes, many people use py for shorthand anyway as it saves them some typing so I think adding that block anyway would be great.

JulienPalard: That's probably not really the responsibility of the readme_renderer tests to validate this, but it may not hurt to add it so we're also ensuring that this particular tag, probably the most used in this context, still work, so I added it.

tests/test_markdown.py
```diff
@@ -0,0 +1,86 @@
...
+def _do_test_with_files(test_name):
+    md_markup = read('{0}.md'.format(test_name))
+    expected_html = read('{0}.html'.format(test_name))
+    out = render(md_markup)
+    assert out == expected_html.rstrip()
+
+
+def read(fn):
+    path = os.path.join(os.path.dirname(__file__), 'fixtures', 'markdown', fn)
+    with io.open(path, encoding='utf-8') as f:
```
AraHaan: No need for using io here either.

tests/test_any.py
```diff
@@ -0,0 +1,45 @@
...
+        'test_rst_006.rst',
+        'test_rst_007.rst'}
+    if os.path.basename(src_filename) in too_short:
+        """sniff won't match in any cases for those files being too short
+        tests to be sniffed correctly (like just an "<a href", or just
+        a <script>alert("Hello")</script>) """
+        return
+    # Get our Markup
+    with io.open(src_filename, encoding='utf-8') as f:
+        markup = f.read()
```
JulienPalard: `open` from Python 2.7 had no encoding parameter, while we want this to work on pypi-legacy, we still need to keep Python 2 compatibility, also that's how tests are written for rst, so it's probably nice to write them the same way for consistency, and as I don't want to modify rst tests in a markdown branch, I think we should keep this that way. I'm not fluent in Python 2 / Python 3 compatible code, but as far as I searched it looks the right way to do it.

tests/test_markdown.py
JulienPalard: Same, Python 2 compatibility, to give the encoding.

AraHaan: Alright, so we can change it later when python 2 is decided to be fully dropped?

JulienPalard: Yes, I'm OK to change it later when python2 compat is dropped.

Look like pypi-legacy has an indeterminate lifespan, but if we pin a legacy version of readme_renderer in pypi-lecacy requirements, we'll be able to drop python2 compatibility. For the moment, I think it's better to keep with python2 compatibility so pypi-legacy can at least get markdown.

AraHaan: Yeah, or port python 2's open function to accept `encoding`` as an paramiter like how python 3 does. I would have to figure out a patch for python 2 then to allow such.

JulienPalard: I'm not sure backporting the `encoding` parameter to Python 2's `open` is worth the effort since Python 2 is end of life. Let's just stick, [like](https://github.com/search?l=Python&q=%22import+io%22&type=Code&utf8=%E2%9C%93) [everyone](http://stackoverflow.com/questions/10971033/backporting-python-3-openencoding-utf-8-to-python-2) [else](http://python-future.org/compatible_idioms.html#file-io-with-open), to a **simple** and **temporary** "import io".

Also, it will be easier to port Python 2 using `io.open` than Python 2 using `open`, as `Python 3`'s open [is](https://docs.python.org/3/library/io.html#io.open) `io.open`.

pradyunsg:
This seems fine to me in principle, barring one thing.

Could you rename "any" to something else (like base)? Because this is a module; naming it to mask a built-in is really, not nice.

JulienPalard:
@pradyunsg done.

tests/test_markdown.py
pradyunsg: A comment here might be useful, mentioning that this a Py2 compatibility thing.

pradyunsg:
Thanks @JulienPalard!

I think this needs @dstufft's review now, which I presume would happen after July 4th. :)

di:
Hi @JulienPalard, thanks for your work here. Since we now have a way for the user to explicitly describe the content-type of their README, being able to infer it isn't necessary. I really appreciate you taking a shot at it though!

JulienPalard:
Hi @di,

Nice job, a "Content-Type" metadata is clearly the right way to do it cleanly. Autodetection will always keep a few degree of errors.

Yet I don't think the 100k packages currently already uploaded will upgrade anytime soon, so fallbacking on autodetection may still make sense when the content type is not provided.

So in one hand, fallbacking to auto-detection when content type is missing fix all already-uploaded packages and permits to omit it when it's not necessary.

On the other hand auto-detecting may have the counter-productive effect of keeping people away from the Content-Type metadata, leading sometimes to surprises (misdetection).

I think I prefer *not* doing auto-detection (only relying on Content-Type defaulting to rst), but it's hard to keep those badly rendered READMEs online.

PR #55, opened 2017-09-05
Markdown

### LINKED ISSUE #1: Markdown support

di:
This PR continues @msabramo's work from #3, specifically:

* Changes `markdown.render` to have the same API as `rst.render`
* Updates the markdown tests to use the same file-based approach as the .rst tests (this gets all the HTML and markdown strings out of the tests)
* Switches to the CommonMark variant, as per https://packaging.python.org/specifications/#description-content-type

Fixes #1, closes #3, closes #51.


readme_renderer/markdown.py
```diff
@@ -0,0 +1,33 @@
...
+from .clean import clean
+
+
+variants = {
+    "CommonMark": commonmark,
+}
+
+
+def render(raw, variant="CommonMark"):
+    renderer = variants.get(variant, commonmark)
```
dstufft: Does it make sense to fall back to CommonMark here if the variant is not known? I'm kind of thinking that it might be better to treat that as an error case, and let callers handle the fallback. Otherwise it isn't possible to determine a variant that doesn't exist.

di: Good call. Addressed.

PR #67, opened 2018-03-18
Add support for GitHub-flavored Markdown

jonparrott:
This uses the [cmarkgfm](https://pypi.org/project/cmarkgfm/) package to render the "gfm" variant of Markdown. It also preferentially uses cmarkgfm for CommonMark as it is CommonMark compliant (and fast).

Notes:
* The cmarkgfm library is currently an *extra* dependency (``readme_renderer[gfm]``) mostly because I don't yet have wheels for cmarkgfm.
* I added a ton of fixtures for gfm based on GitHub's [own spec](https://github.github.com/gfm/). They had to be tweaked a little because bleach changes the output very slightly (such as adding `rel="nofollow"`).


jonparrott:
Context: https://github.com/pypa/packaging-problems/issues/126

di:
Given how many folks will probably want GFM over CommonMark, I wonder if we should make that the default variant.

pradyunsg:
> Given how many folks will probably want GFM over CommonMark, I wonder if we should make that the default variant.

+1

jonparrott:
> Given how many folks will probably want GFM over CommonMark, I wonder if we should make that the default variant.

Baby steps. Let's get this in first. Solid next steps after that are:

1. Get wheels built for cmarkgfm.
1. Transfer cmarkgfm over to pypa.
1. Make cmarkgfm a non-optional dependency here, remove CommonMark (the package) dependency.

In parallel:

1. Update warehouse to support the `gfm` variant of Markdown.

tests/test_markdown.py
```diff
@@ -2,36 +2,44 @@
...
-        md_markup = f.read()
+        md_markup = f.read().strip()
 
     # Get our expected
     with io.open(html_filename, encoding="utf-8") as f:
-        expected = f.read()
+        expected = f.read().strip()
 
-    assert render(md_markup, variant=variant) == expected
+    assert render(md_markup, variant=variant).strip() == expected
```
di: Why is `.strip()` necessary here? Can we just add/remove newlines from the fixtures as necessary?

di:
Ah yeah, I meant to make that comment over at https://github.com/pypa/packaging-problems/issues/126 instead. Moving...

tests/test_markdown.py
jonparrott: Sure. The reason I added this is because `bleach` can occasionally consume the trailing newline from things- especially when it fixes broken HTML. This is mostly just affecting on particular fixture, so I can likely just do that. WDYT?

di:
> Update warehouse to support the `gfm` variant of Markdown.

This should Just Work™, the only change to Warehouse necessary would be bumping the `readme_renderer` dependency.

jonparrott:
@di see https://github.com/pypa/packaging-problems/issues/126#issuecomment-374307401

tests/test_markdown.py
jonparrott: done in 93589c7

di:
Thanks @jonparrott! I think we should probably hold off on a release until we do `#3`:

> Make cmarkgfm a non-optional dependency here, remove CommonMark (the package) dependency.

jonparrott:
That's totally fine. I'm trying to get wheels built now. If you know anyone who can help with Windows builds I'd love to chat.

PR #68, opened 2018-03-20
Switch to cmarkgfm completely for rendering Markdown

jonparrott:
Follow up to #67.

Towards https://github.com/pypa/packaging-problems/issues/126

~~Note: There are [wheels](https://github.com/jonparrott/cmarkgfm-wheels) for `cmarkgfm` for basically everything but Python 2.7 on Windows. This is due to the hilariously ancient version of MSVC needed to compile the wheel and `cmark` just outright being uncompilable as-is. Basically, users of 2.7 on Windows will *not* be able to use `readme_renderer` at all. As long as we're okay with that, this can merge.~~

Wheels are available across Windows, Linux, and Mac for 2.7, 3.4, 3.5, and 3.6.

### DIFF
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -13,25 +13,15 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
 
-from CommonMark import commonmark
+import cmarkgfm
 
 from .clean import clean
 
 
-try:
-    import cmarkgfm
-except ImportError:
-    cmarkgfm = None
-
-
-variants = {}
-
-if cmarkgfm is not None:
-    variants["gfm"] = cmarkgfm.github_flavored_markdown_to_html
-    # Preferentially use cmarkgfm for CommonMark.
-    variants["CommonMark"] = cmarkgfm.markdown_to_html
-else:
-    variants["CommonMark"] = commonmark
+variants = {
+    "gfm": cmarkgfm.github_flavored_markdown_to_html,
+    "CommonMark": cmarkgfm.markdown_to_html,
+}
 
 
 def render(raw, variant="CommonMark", **kwargs):
--- a/readme_renderer/txt.py
+++ b/readme_renderer/txt.py
@@ -13,14 +13,7 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
 
-try:
-    from html import escape as html_escape
-except ImportError:
-    from cgi import escape as _html_escape
-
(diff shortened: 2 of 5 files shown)


jonparrott:
@di alternatives for Windows 2.7 users include:

* Use a platform specifier for `cmarkgfm` and then warn them if they try to use it. (install succeeds, usage fails)
* Use a platform specifier for `cmarkgfm`, keep `CommonMark` as a fallback, work for `CommonMark` variant but fail for `gfm` varianet.

jonparrott:
I'm a bit baffled as to how this change caused the `txt` tests to fail on `2.7`. Weird.

jonparrott:
> I'm a bit baffled as to how this change caused the txt tests to fail on 2.7. Weird.

Figured it out. `CommonMark` was pulling in `future` which cleverly provides `html.escape` which means that the branch in `txt.py` that checks for `html.escape` and falls back to `cgi.escape` was previously never used. Removing `CommonMark` removed `future` and thus `html` and made the previously unused branch trigger for `2.7` - which has different escape behavior for single quotes!

Fixed in 5e7c7d1 by just including `future` as a direct dependency and removing the problematic branch.

anthrotype:
@jonparrott have you tried compiling the Windows 2.7 wheels using mingwpy toolchain?
https://anaconda.org/carlkl/mingwpy
https://github.com/mingwpy/mingwpy
https://mingwpy.github.io/

jonparrott:
@anthrotype nope, is there an easy way to do that with `multibuild`? I'm using that over in https://github.com/jonparrott/cmarkgfm-wheels.

jonparrott:
Aww it seems [mingwpy is abandoned](https://groups.google.com/forum/#!topic/mingwpy/1k_BLFPLmBI)?

anthrotype:
> Aww it seems mingwpy is abandoned?

yes, but it still works for 2.7. 
you can `pip install -i https://pypi.anaconda.org/carlkl/simple/ mingwpy`

I use it like this in another project where I need c++11 stuff not available for MSVC2008
https://github.com/googlei18n/compreffor/blob/master/.appveyor.yml
https://github.com/googlei18n/compreffor/blob/master/tox.ini#L4

jonparrott:
@anthrotype ah great. I've give that a shot this afternoon. Thank you!

jonparrott:
Thank you for the help @anthrotype! I got wheels working for Python 2.7 on Windows!

@di this should be good to go. :)

PR #72, opened 2018-03-30
Release 18.0

jonparrott:


jonparrott:
> Did we decide not to change the default variant to GFM?

Maybe I missed something, but I don't remember that ever being a question. I would say let's get this out there and tested before doing something like that.

di:
I think it probably makes sense to make GFM the default variant at the same time that we introduce multiple variants.

Judging by the conversations/issues I've seen, it seems like most people using `text/markdown` now actually want GFM. Making the default variant GFM now means improves the UX: folks that currently have semi-broken Markdown descriptions will magically get them fixed, and they won't need to change their `setup.py` to specify the GFM variant in future releases, only to be told one day that it's now unnecessary because we've changed the default.

Also since GFM is a strict and optional superset of CommonMark, there won't be any broken descriptions as a result of us changing the default. IIRC the only reason we have the CommonMark variant is because it was our only solid option at the time, now that we have GFM, I can't imagine why someone would actually specify the CommonMark variant (unless they had some moral or ideological opposition to GFM).

And, if changing the default breaks something that we missed when testing this, we'll figure it out pretty quick, and the worst thing that would happen is that some releases are temporarily broken on a site that's in beta anyways 😉

jonparrott:
You've got me convinced. Hold off on this PR until while I send another to make it so.

jonparrott:
Once https://github.com/pypa/readme_renderer/pull/73 is in, I'll rebase this.

jonparrott:
Rebased and changelog updated. Should be good to go now @di and @dstufft. :)

PR #74, opened 2018-04-01
Change Github-flavored Markdown identifier to 'GFM'

di:
Per https://tools.ietf.org/html/rfc7764#section-3.2

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Changes
 =======
 
+18.1 (2018-04-01)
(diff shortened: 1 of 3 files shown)


di:
I think case-insensitivity is not worth the trouble, considering this will be rarely set, and handling case-insensitivity will have to travel downstream as well to where these variants are being validated by Warehouse.

jonparrott:
Sounds good for now. :) We can always revisit if folks run into it.

PR #77, opened 2018-04-03
Allow details and summary tags

### LINKED ISSUE #76: Allow `details` and `summary` tags
From: https://github.com/pypa/warehouse/issues/3539:

>Thanks for adding markdown support!
> 
>I just tried uploading one of my projects with the ``long_description_content_type='text/markdown'``. Overall it looks good (apart for the relative links, which I will need to fix manually). However, I noticed that the ``<details>`` tags aren't being rendered correctly. This might have to do with the quote blocks that I put inside of the HTML tags, but the page renders correctly on Github.
>
>Reference (see the ``FAQ`` section at the bottom):
>
>https://test.pypi.org/project/rtv/1.22.1.post1/
>https://github.com/michael-lazar/rtv/blob/master/README.md#faq

This should be simple as adding the `details` and `summary` to `ALLOWED_TAGS`.

pradyunsg:
Fixes #76

### DIFF
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -27,9 +27,10 @@ ALLOWED_TAGS = [
     "strong", "ul",
 
     # Custom Additions
-    "br", "cite", "col", "colgroup", "dd", "del", "div", "dl", "dt", "h1",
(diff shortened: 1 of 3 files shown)


di:
@pradyunsg Would you mind adding a test for this? Also looks like there are linting errors.

Issue #78, opened 2018-04-03
GFM table: Optional whitespace next to pipes


nnadeau:
# Expected
The following table without spaces between the pipes is valid GitHub Flavoured Markdown, according to the following parsers:
- https://atom.io/packages/markdown-preview
- This issue's input box 

```
|Foo|Bar|
|---|---|
|FooFoo|BarBar|
```

|Foo|Bar|
|---|---|
|FooFoo|BarBar|

# Actual
```
python3 setup.py check -r -s 

warning: Check: :1: (ERROR/3) Undefined substitution referenced: "Foo|Bar".

warning: Check: Invalid markup which will not be rendered on PyPI.

error: Please correct your package.
```

# Solution
- Adding spaces next to all the pipes stops the errors:
```
| Foo | Bar |
| --- | --- |
| FooFoo | BarBar |
```

| Foo | Bar |
| --- | --- |
| FooFoo | BarBar |

- This issue probably stems from how the [GFM spec was interpreted](https://github.github.com/gfm/#tables-extension-), specifically: `[...] Spaces between pipes and cell content are trimmed [...]
  - Spaces are optional

di:
Have you correctly set the `long_description_content_type` in your `setup.py`? The "Undefined substitution referenced" error is from the rST checker, which shouldn't be used here.

nnadeau:
`long_description_content_type='text/markdown',`, just like in the two example repos: 
- https://github.com/jonparrott/gfm-markdown-description-example/blob/master/setup.py
- https://github.com/di/markdown-description-example/blob/master/setup.py

nnadeau:
also:
```
Name: readme-renderer
Version: 17.4

Name: setuptools
Version: 39.0.1
```

jonparrott:
You'll need to update to `readme-renderer >= 18.0`. We added all the fun markdown stuff in `18.0`.

nnadeau:
```
Name: readme-renderer
Version: 18.1
```

Gives me this:

```
python3 setup.py check -r -s            
                                                                                                                    
warning: Check: :1: (ERROR/3) Undefined substitution referenced: "Foo|Bar".

warning: Check: The project's long_description has invalid markup which will not be rendered on PyPI.

error: Please correct your package.
```

di:
Sorry, I totally forgot how the distutils integration works. The `-r` flag is for `--restructuredtext`:

```
--restructuredtext (-r)  Checks if long string meta-data syntax are
[...]

PR #80, opened 2018-04-03
Update docs about check to talk about Markdown

di:
Per #78.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,8 +12,11 @@ reStructuredText (``.rst``), and plain text.
 Check Description Locally
 -------------------------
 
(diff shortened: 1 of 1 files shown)


jonparrott:
Should we also consider throwing an outright error if `-r` is specified and the content-type is set to markdown? (We can do it in a separate bug/PR, just wanted to bring it up)

di:
@jonparrott That'd be great, but I don't think distutils provides a way to get arbitrary metadata fields, and it doesn't support `long_description_content_type` field yet either, so I'm not sure it would be possible.

PR #82, opened 2018-04-06
Add Pygments-based syntax highlighting for Markdown

### LINKED ISSUE #75: Support Syntax Highlighting w/ GFM
From https://github.com/pypa/warehouse/issues/3508:

> GItHub has syntax highlighting for code blocks, PyPI has not.

jonparrott:
Resolves #75


jonparrott:
Note: I verified that this is more or less identical to the same output for `rst` both by testing similar docs in both formats and grafting rendered test cases into warehouse's markup. It seems to be good.

Issue #63, opened 2018-02-01
feature: report success explicitly


brainwane:
Right now, when a check is successful:

~~~
$ python setup.py check -r -s
running check
$ 
~~~

It would be good to explicitly say something like "Check successful".

PR #83, opened 2018-04-10
Warn about markdown content type when checking RST

### LINKED ISSUE #81: Investigate whether we can raise an error if checking a Markdown description
In #80 @jonparrott said:

> Should we also consider throwing an outright error if -r is specified and the content-type is set to markdown? (We can do it in a separate bug/PR, just wanted to bring it up)

I said:

> That'd be great, but I don't think distutils provides a way to get arbitrary metadata fields, and it doesn't support `long_description_content_type` field yet either, so I'm not sure it would be possible.

It'd be great to investigate whether this is true or not.

It's worth mentioning here that distutils is unlikely to support anything beyond Metadata 1.1 (and we also need to be backwards-compatible with older versions of distutils).

jonparrott:
Closes #81 and consequently #63.

### DIFF
--- a/readme_renderer/integration/distutils.py
+++ b/readme_renderer/integration/distutils.py
@@ -13,8 +13,10 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
 
+import cgi
 import io
 
+import distutils.log
 from distutils.command.check import check as _check
 
 from ..rst import render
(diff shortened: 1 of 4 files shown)


jonparrott:
Merging so I can prepare a release, but happy to address any concerns post-hoc before merging the actual release.

readme_renderer/integration/distutils.py
```diff
@@ -26,6 +28,16 @@ def check_restructuredtext(self):
         Checks if the long string fields are reST-compliant.
         """
         data = self.distribution.get_long_description()
+        content_type = getattr(
+            self.distribution.metadata, 'long_description_content_type', None)
```
di: I'm not sure when `self.distribution.metadata` came to be -- is this available in all version of distutils? Otherwise we might want to test whether these attributes exist.

jonparrott: It's present in Python 2.7 [here](https://github.com/python/cpython/blob/77f0a41d72886513f5e9277e5587455745c29ac1/Lib/distutils/dist.py#L151), and going back quite a bit it looks like it's been around for at least [18 years](https://github.com/python/cpython/blame/39d59b47b70d557e0aec5c8f35ede0a7d73f4edc/Lib/distutils/dist.py#L121).

di: That's probably far enough 😉

PR #97, opened 2018-04-20
Make Python 3 the default lexer for highlighting Python code in Markdown

### LINKED ISSUE #96: Python 3.5 new keywords, async and await, not syntax highlighted
From https://github.com/pypa/warehouse/issues/3770:

> See for instance https://pypi.org/project/influxdbpusher/
>
> It has code snippets  with `async def` and `await`, which are not being properly highlighted as keywords.

jonparrott:
Fixes #96

### DIFF
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -24,6 +24,15 @@ from six.moves import html_parser
 from .clean import clean
 
 
+# Make code fences with `python` as the language default to highlighting as
+# Python 3. Add an extra alias for python2 to map to the Python 2 lexer
+# (`python`) so that users can explicitly use Python 2 highlighting.
+_LANG_ALIASES = {
+    'python2': 'python',
+    'python': 'python3',
+}
+
+
 variants = {
     "GFM": cmarkgfm.github_flavored_markdown_to_html,
     "CommonMark": cmarkgfm.markdown_to_html,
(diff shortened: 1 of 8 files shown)


di:
I'm not sure we want to introduce a shortname (`python2`) that doesn't actually exist in pygments. I'm also hesitant to add anything that produces inconsistencies between how these READMEs are rendered on PyPI and how they're rendered elsewhere (e.g. on Github), which is what will happen if folks use the `python` shortname with this PR.

Perhaps the original issue is actually invalid, and we should tell users that want to highlight Python 3 code that they should use the `python3` shortname? I agree that `python` should probably map to the Python 3 lexer _someday_, but I'm not sure if it's for us to decide.

jonparrott:
Yeah I'm hestant here, but I can think of some arguments *in favor* of doing this.

The biggest is that GitHub's highlighter defaults `python` to `python3` for example this is just `python` after the code fence:

```python
async def meep():
   pass

await meep()
```

Since the goal of supporting `gfm` is to be as close to github as possible, I am okay with us aliasing `python` to `python3`. Especially because without aliasing there's no way to get highlighting to work correctly for Python 3 code on both github and warehouse at the same time.

Whether or not we should allow `python2` to work is another story. I wanted at least *some* sort of escape hatch in case some user really, really needed Python 2 highlighting mode. But, I'm happy to leave that out until someone asks for it.

di:
Ah, I didn't realize GFM was already mapping `python`->`python3`. In that case, we probably should as well.

jonparrott:
> Let's ditch the python2 shortname.

Done.

PR #99, opened 2018-04-20
Include docinfo when rendering rst

### LINKED ISSUE #61: Include docinfo table
I'm used to put standard docinfo in my README.rst and thus in package's description.

The `rst.render()` function considers only the `fragment`, though, so that information does not appear on PyPI/Warehouse.

To trick it I indent the docinfo fields, so that they are rendered as a `<table>` within a `<blockquote>`: that works, but the result is somewhat ugly.

What do you think about prepending the `docinfo` part, if present?

theacodes:
Closes #61.

Screenshot of docinfo on Warehouse:

screen shot 2018-04-20 at 10 47 51 am

### DIFF
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -115,7 +115,7 @@ def render(raw, stream=None, **kwargs):
     except SystemMessage:
         rendered = None
     else:
-        rendered = parts.get("fragment")
+        rendered = parts.get("docinfo", "") + parts.get("fragment", "")
 
     if rendered:
         return clean(rendered)
--- a/tests/fixtures/test_rst_docinfo.html
+++ b/tests/fixtures/test_rst_docinfo.html
@@ -1,0 +1,19 @@
+<table>
+<colgroup><col>
+<col>
+</colgroup><tbody>
+<tr><th>Project:</th><td>pg_query – Pythonic wrapper around libpg_query</td>
+</tr>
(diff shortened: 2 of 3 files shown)


di:
Would it make sense to put the docinfo at the _end_ of the rendered description, rather than at the beginning?

jonparrott:
Possibly, but we'd be breaking precedence with literally every other rst renderer (docutils, sphinx, and github).

di:
Well, we already are by only rendering the fragment. My concern is that most users reading a projects description probably want to read the first couple lines of the description, not all this metadata. We've already gotten some complaints that there is not enough information density on small screens, and I think this might aggravate it.

I think it might also be unclear to users that this metadata is coming from the description, and not from PyPI itself. For example, in your screenshot, the date of the "created" line is pretty much always going to disagree with the release date. At least in this example, a lot of this information (license, author, project name) is already going to be included in the sidebar on PyPI anyways as well.

I'd like to hear what others (@dstufft, @ewdurbin) think.

jonparrott:
Totally fair points. Definitely interested to see what others say.

theacodes:
After discussing with @dstufft and @di in person, we decided to merge this as-is. Although this isn't pretty, we don't want to deviate from how GitHub renders things. Additionally, this is optional and present on a fairly small number of readmes.

Issue #52, opened 2017-07-03
Could the Distutils check generate an HTML document ?


Lucas-C:
Hi,

I tested it and this would just be a matter of adding the following at the end of https://github.com/pypa/readme_renderer/blob/master/readme_renderer/integration/distutils.py :
```
else:
    with open('long_description.html', 'w') as long_desc_file:
        long_desc_file.write(markup)
```

Now, are you OK to add this feature ?
I find it very useful to preview the description before uploading a package.

dgasaway:
I concur. Making this change and previewing the HTML helped me find a few issues in my readme.

di:
I think it might make more sense to add a separate command just to generate the HTML. #53 is an attempt at this, but is unfinished.

PR #101, opened 2018-05-14
Add `render_readme` command to render HTML locally.

khorn:
This allows users to render HTML locally in order to preview the
rendered markup prior to uploading a package.  Optionally, the
user can also control syntax highlighting and send the markup to
a web browser.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,33 @@ your Markdown description as Markdown if it has some invalid or incorrect
 syntax.
 
 
+Render Markup Locally
+---------------------
+
+You can render the HTML that will be generated from your ``long_description`` using:
+
+.. code-block:: console
+
+	$ python setup.py render_readme
+
+The rendered HTML will be output to the console by default.
+
+If you would like to send the output to a web browser instead, use the ``--preview`` flag:
+
+.. code-block:: console
+
+	$ python setup.py render_readme --preview
+
+You can also control the Pygments style used in syntax highlighting using the
+``--style`` flag, or eliminate syntax highlighting altogether by using the
+``--no-color`` flag:
+
+.. code-block:: console
+
+	$ python setup.py render_readme --style=monokai # <-- Pygments style name
+	$ python setup.py render_readme --no-color
+
+
 Code of Conduct
 ---------------
 
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ You can render the HTML that will be generated from your ``long_description`` us
 
 .. code-block:: console
 
-	$ python setup.py render_readme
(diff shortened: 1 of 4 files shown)


khorn:
FYI, replaces #53

theacodes:
@khorn it's a good start, but there seems to still be some todos. Will you ping me when you're ready for me to review (or if you get stuck?)

khorn:
@theacodes Ping.

theacodes:
Hi @khorn, I don't think it's very useful to have console output for this. I would prefer us just to do the HTML preview.

hsiaoyi0504:
Is there any update of this ? I agree that having console output for this doesn't quite useful, but it would be good to have a output html generated for more easier sharing.

theacodes:
@hsiaoyi0504 it's up to @khorn if they want to update their PR. You're also welcome to start a new PR based on this one to address review comments.

khorn:
@theacodes Sorry it's been so long...

I'm a bit confused by your comment.  If the intention was to only do an HTML preview in a web browser, then I don't understand why #53 wasn't accepted.  It does exactly that and works (or at least worked...it appears to have developed some conflicts) perfectly well.

The intent of this change was to (a) replace/update that branch, and (b) also address the request in #52, which asks for a command to generate an HTML preview in a file.

This was exactly what I was asked to do at the PyCon sprint.  The only difference is I decided to send the generated HTML to stdout, rather than a file, as that seemed more flexible.

I figured that sending the results to stdout would allow the command to be sent to linters, spell checkers, etc. perhaps as part of a release checklist or CI/CD workflow.

Personally, it seems to me that stdout output is far *more* useful than a webbrowser preview.  I would certainly rather have a tool that sent HTML to stdout, which I could then pipe anywhere, rather one that only allowed me to preview the HTML in a browser.

Why do you disagree?

theacodes:
@khorn sorry for the confusion.

I think it's surprising behavior for an interactive command-line tool to output non-human-readable code. I'd prefer if we had the following behavior, let me know what you think:

1. Running `render-readme` with no arguments produces a `readme.html` file in the current directory.
2. Running `render-readme --output -` outputs to stdout. (this behavior is consistent with [argparse](https://docs.python.org/3/library/argparse.html#filetype-objects)

For (2) don't bother doing console color highlighting, as the use case for outputting the raw html to stdout is to pipe to other programs which likely won't be fans of any escape sequences.

khorn:
@theacodes I'm fine with your suggestions, just haven't gotten to it yet.  I do still intend to!

theacodes:
Cool! No worries. :)

PR #103, opened 2018-05-25
Add specific tests for the raw directive

theacodes:


theacodes:
@di FYI, tests-only change, merging.

Issue #112, opened 2018-07-28
Inconsistencies and problems in when images are rendered from RST


mila:
Investigating https://github.com/pypa/warehouse/issues/4023, I found multiple related problems:

- HTML code for SVG images can be quite different from other image types.
- For SVG, width and height are assigned to width="..." and height="..." attributes, so relative dimensions cannot work.
- For non-SVG image types, width and height are ignored.
- For SVG images, align="..." attribute is set, which does not help with styling.
- For non-SVG images, align is ignored.
- For most images, alt="" fallbacks to image URL, not for SVG.
- Detection of SVG is broken, e.g. URL like `.../django-multi-gtfs.svg?branch=master` is not detected as SVG.

di:
@mila I think this is fixed now that #113 and #114 are merged and released (https://pypi.org/project/readme_renderer/22.0/), can you close if that's accurate, or update this with anything that's currently outstanding?

mila:
It would be great to deploy Warehouse using the new version. The goal is to close https://github.com/pypa/warehouse/issues/4023. 

Unless I missed something, there should be no more work in readme_renderer. So, I'm closing this issue.

dstufft:
I'm reopening this, because the fix in https://github.com/pypa/readme_renderer/pull/114 isn't going to work on PyPI due to our restrictive CSP policy which disallows inline styles. We're not going to relaxing that, so in order for this to actually work, readme_renderer will need to emit css classes in all cases.

mila:
Good catch. I'm afraid that it's impossible to define css classes for each possible width or height. I see two options:

1) Do not support setting of width or height.  `class` should be removed from allowed attributes to prevent CSP violations.

2) Patch `rst` renderer to generate `width="..."` and `height="..."` instead of `style="..."`. Only absolute (`px`) dimensions would be supported.

There should be no problem with image alignment, it's already implemented using css classes.

dstufft:
I guess I'd say that (2) is probably the best option here, unless we think that absolute dimension part is no go, then maybe (1).

Although maybe we could do something like (2), but also add a couple common sizes for relative dimensions that people can use too? I'm not sure how needed these things are TBH, so I don't know if either of these things are worth the effort or not.

I guess we should probably also ping @nlhkabu for input on whether mandating absolute sizing is going to break the UX or not.

mila:
(1) has to be done in any case and #121 should make the job. 

[...]

PR #113, opened 2018-07-28
Unify handling of SVG and other images in RST.

mila:
ReadMeHTMLTranslator was overidden to not to render object tags for SVG.
Unfortunately it produced different results than for PNG or other
image types.

This a first pull request, which  addresses #112. Once image handling is unified, other issues 
can can fixed.


theacodes:
Hmm, based on https://sourceforge.net/p/docutils/bugs/247/, docutils should already be outputting img tags for svgs. Can you help me understand why we need `object_image_types = {'.swf': 'application/x-shockwave-flash'}`?

mila:
I have docutils 0.14 installed and they still render `<object>` for svg.

`ReadMeHTMLTranslator` extends `docutils.writers.html4css1.HTMLTranslator` which extends `docutils.writers._html_base.HTMLTranslator`.  

 - `_html_base.HTMLTranslator` sets `object_image_types = {'.swf': 'application/x-shockwave-flash'}`
 - `html4css1.HTMLTranslator` adds svg there `{'.svg': 'image/svg+xml',  '.swf': 'application/x-shockwave-flash'}` with a comment that _"html4css1 strives for IE6 compatibility"_
 - I removed `.svg` and kept `.swf` there, just to make the change minimal.

theacodes:
Gotcha. Could we use a newer base translator?

mila:
> Gotcha. Could we use a newer base translator?

I'm not sure about that. There are two implementations:  `html4css1` and `html5_polyglot`. 
The first one is the one we are using now, the latter claims to "_generates polyglot markup: HTML5 that is also valid XML._"

The polyglot markup may be a better choice, but it generates different output and makes tests failing.
So if the translator has to be changes, I would do that in a separate issue.

theacodes:
Sounds pretty reasonable, we should do that in a follow-up.

I'd like @di to take a look at this as well, as anything that affects the sandboxed output here should be double-checked.

readme_renderer/rst.py
```diff
@@ -25,26 +24,9 @@
...
-                    if attribute == "uri":
-                        atts["src"] = value
-                    else:
-                        atts[attribute] = value
-
-            # toss off `object` tag
-            self.body.pop()
-            # add on `img` with attributes
-            self.body.append(self.starttag(node, "img", **atts))
+    # We need to swap RST's use of `object` with `img` tags
```
theacodes: Can you update this comment to mention that we're override the base class to *not* output svg with `object`?

mila: Done.

readme_renderer/rst.py
```diff
@@ -25,26 +24,8 @@
...
-                        atts["src"] = value
-                    else:
-                        atts[attribute] = value
-
-            # toss off `object` tag
-            self.body.pop()
-            # add on `img` with attributes
-            self.body.append(self.starttag(node, "img", **atts))
+    # Overrides base class not to output `<object>` tag for SVG images.
+    object_image_types = {'.swf': 'application/x-shockwave-flash'}
```
di: Let's just make this `object_image_types = {}`, we don't actually want to support _any_ `<object>` tags at all.

tests/fixtures/test_rst_svg_attrs.html
```diff
@@ -0,0 +1 @@
+<img alt="alternate text" src="https://example.com/badge.svg">
```
di: We're kind of experiencing a regression here, right? Before, this:

```rst
.. image:: https://example.com/badge.svg
   :height: 100px
   :width: 25.0%
   :alt: alternate text
   :align: right
```

would become:

```html
<img align="right" alt="alternate text" height="100px" src="https://example.com/badge.svg" width="25.0%">
```

but now it becomes:

```html
<img alt="alternate text" src="https://example.com/badge.svg">
```

mila: Yes, but the current behavior is problematic:
 - `align="right"` does not actually align the image,  [CSS](https://github.com/pypa/warehouse/pull/4391/files) is necessary for that.
 - `height="100px"` wont have any effect, because units (`px`) must be added to CSS, not image attributes
 - Those attributes are preserved for SVG images only (or more precisely for URLs ending ".svg")

https://github.com/pypa/readme_renderer/pull/114 addresses image alignment and dimensions. Without unifying SVG with other image types, the subsequent PR would fix styling of non-SVG images only.

(I created two PRs because the second relaxes what HTML is allowed, so there can be more discussion.)

readme_renderer/rst.py
mila: Done.

tests/fixtures/test_rst_svg_attrs.html
di: Got it, thanks for the explanation.

theacodes:
Thank you, @mila!

PR #114, opened 2018-07-28
Width and alignment of RST images

mila:
RST renders images with `class="align-..." style="width: ...; height: ..."`.  This pull request stops bleach from removing image styling. 

Allowing `class` and `style` attributes can be controversial, but this PR does not introduce anything new - `width` and `height` attributes are already allowed (and other styles won't be allowed), and `class` can be already assigned to `span` and other elements.

This should follows PR https://github.com/pypa/readme_renderer/pull/113 (merge it first for a smaller diff) and fixes issue https://github.com/pypa/readme_renderer/issues/112.  
Overall goal is to fix https://github.com/pypa/warehouse/issues/4023.

### DIFF
--- a/tests/fixtures/test_GFM_style.html
+++ b/tests/fixtures/test_GFM_style.html
@@ -1,0 +1,1 @@
+<img src="https://example.com/badge.png" style="width: 20px; height: 20px;">
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -42,7 +42,7 @@ ALLOWED_ATTRIBUTES = {
     # Custom Additions
     "*": ["id"],
     "hr": ["class"],
-    "img": ["src", "width", "height", "alt", "align"],
+    "img": ["src", "width", "height", "alt", "align", "class"],
     "span": ["class"],
     "th": ["align"],
     "td": ["align"],
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -42,7 +42,7 @@ ALLOWED_ATTRIBUTES = {
     # Custom Additions
     "*": ["id"],
     "hr": ["class"],
-    "img": ["src", "width", "height", "alt", "align", "class"],
+    "img": ["src", "width", "height", "alt", "align", "class", "style"],
     "span": ["class"],
     "th": ["align"],
     "td": ["align"],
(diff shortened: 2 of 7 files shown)


theacodes:
@mila can you rebase this?

mila:
> @mila can you rebase this?
Rebased.

westurner:
"CSS injection: what's the worst that can happen?" https://stackoverflow.com/questions/718611/css-injection-whats-the-worst-that-can-happen

> Injection of arbitrary CSS can lead to javascript execution.


https://bleach.readthedocs.io/en/latest/clean.html#allowed-styles-styles

```python
ALLOWED_STYLES = [
    "width", "height",
]
```

Does this set ``bleach.sanitizer.ALLOWED_STYLES`` somewhere else or just the global constant ALLOWED_STYLES?

westurner:
@dstufft Would having a separate domain for admin / logged-in actions like admin.pypi.org minimize the risk of things such as XSS in rendered long_descriptions?

westurner:
Thanks.

While I think bleach is a good package, and I'd add tests to bleach if I
was aware of any untested XSS issues, I still think it best to run
user-supplied markup on a different domain; similar to how GitHub.io and
ReadTheDocs.io work; though a subdomain such as admin/edit.pypi.org instead
of a totally different TLD may actually be fine?

“Stop using .IO Domain Names for Production Traffic” @nickparsons
https://hackernoon.com/stop-using-io-domain-names-for-production-traffic-b6aa17eeac20

dstufft:
I don't think that GitHub puts user supplied markup on a different domain.. they are putting it on all of these issues, README files, etc. They moved raw files to another domain, which PyPI already does.

westurner:
AFAIU, user-supplied JS (which may be present in HTML renders of
lightweight markup languages, static HTML, Jupyter notebooks,) is on a
separate domain in order to limit the impact of potential XSS vulns.

tests/fixtures/test_GFM_style.html
```diff
@@ -0,0 +1 @@
+<img src="https://example.com/badge.png" style="width: 20px; height: 20px;">
```
dstufft: Sorry I hadn't looked at the code changes before, this isn't going to work on PyPI, we don't allow inline styles via CSP, so anything that relies on the ``style`` attribute is not going to work, and I would be a hard -1 on relaxing our CSP rules to allow it.

di:
For future reference, this was reverted in https://github.com/pypa/readme_renderer/pull/121.

PR #120, opened 2018-09-18
Whitelist the allowed CSS classes

dstufft:
This should prevent people from being able to apply classes to their rendered text that we did not otherwise want to allow them to do. For instance they won't be able to apply the same CSS class to their image that our logo uses, which would theoretically let them replace our logo.

### DIFF
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -63,12 +63,10 @@ ALLOWED_CLASSES = {
 
 
 class _CSSClassFilter(html5lib.filters.base.Filter):
-    def __init__(self, *args, allowed_classes=None, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, *args, **kwargs):
+        self.allowed_classes = kwargs.pop("allowed_classes", {})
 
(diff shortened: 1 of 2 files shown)


readme_renderer/clean.py
```diff
@@ -54,21 +56,75 @@
     "width", "height",
 ]
 
+ALLOWED_CLASSES = {
+    "img": ["align-left", "align-center", "align-right"],
+    "span": [c for c in pygments.token.STANDARD_TYPES.values() if c],
```
dstufft: Two things:

* I'm not sure if there is any use of a class inside of a span that isn't there to enable pygments, or if the list of pygments classes is all we need to worry about here.
* @theacodes has set the ``<code>`` element to allow classes in the PR that added syntax highlighting, but I can't determine what classes are valid for the ``<code>`` element. Trying different combinations of things doesn't seem to actually generate code that emits a class on the ``<code>`` element. If we can't figure out how to make a README emit a ``<code>`` element with css classes, we should probably emit ``class`` from ``<code>``.

PR #122, opened 2018-09-21
Recommend using `twine check` instead

di:
This PR updates the README for this repo to recommend using the new `twine check` command instead of `setup.py check -r -s`.

It also adds a deprecation notice for the old command.

This should be merged once the new version of Twine is released.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,30 +12,8 @@ reStructuredText (``.rst``), and plain text.
 Check Description Locally
 -------------------------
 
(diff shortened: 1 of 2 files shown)


ionelmc:
Why exactly is this deprecated?

merwok:
Probably because not everyone is using setup.py, so it’s becoming less central to the workflow (somewhat like how tox replaced setup.py test).

PR #134, opened 2018-10-25
Factor out python-future

### LINKED ISSUE #128: Remove dependency on `python-future`?
python-future is notorious for breaking modules which use `try: / except:` to import version specific modules: (a few examples from things I've encounted: [1](https://github.com/jmcgeheeiv/pyfakefs/pull/441#issue-221957790) [2](https://github.com/Yelp/yelp_clog/pull/18#issue-64272423))

let me know if you'd be interested in me making a PR along these lines -- it seems `future` is only being used for `html.escape` right now:

```diff
diff --git a/readme_renderer/txt.py b/readme_renderer/txt.py
index 0c0c533..3218922 100644
--- a/readme_renderer/txt.py
+++ b/readme_renderer/txt.py
@@ -13,10 +13,18 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
 
-from html import escape as html_escape
+import sys
 
 from .clean import clean
 
+if sys.version_info >= (3,):
+    from html import escape as html_escape
+else:
+    from cgi import escape
+
+    def html_escape(s):
+        return escape(s, quote=True).replace("'", '&#x27;')
+
 
 def render(raw, **kwargs):
     rendered = html_escape(raw).replace("\n", "<br>")
diff --git a/setup.py b/setup.py
index 5f2b64a..16fa18c 100644
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,6 @@ setuptools.setup(
         "bleach>=2.1.0",
         "cmarkgfm>=0.2.0",
         "docutils>=0.13.1",
-        "future",
         "Pygments",
         "six",
     ],
```

asottile:
Resolves #128

### DIFF
--- a/readme_renderer/txt.py
+++ b/readme_renderer/txt.py
@@ -13,9 +13,17 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
 
-from html import escape as html_escape
+import sys
 
 from .clean import clean
+
+if sys.version_info >= (3,):
+    from html import escape as html_escape
+else:
+    from cgi import escape
+
+    def html_escape(s):
+        return escape(s, quote=True).replace("'", '&#x27;')
 
 
 def render(raw, **kwargs):
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,6 @@ setuptools.setup(
     install_requires=[
         "bleach>=2.1.0",
         "docutils>=0.13.1",
-        "future",
         "Pygments",
         "six",
     ],


theacodes:
Thank you!

asottile:
and thank you! :tada:

ssbarnea:
@asottile Can you please make a new relase on pypy as the unlisted html dependency broke few openstack projects and it would be easier if we don't have to merge a workaround.

asottile:
@ssbarnea heh, yet another project I don't have merge access too -- I'm curious though, what broke from the openstack side? (do you have a stacktrace?)

di:
@ssbarnea Could you open a new issue? [The `html` module is in the standard library](https://docs.python.org/3/library/html.html) but it looks like it (or possibly just the `escape` function) was added in 3.2, so the `version_info` check here might be wrong.

Issue #135, opened 2018-10-26
Officially support Python 3.7


di:
- [ ] Test against 3.7
- [ ] Add Python 3.7 classifier

pganssle:
Note that for adding Python 3.7 tests on Travis, you need to use the `xenial` distribution, see, e.g. [this PR](https://github.com/pypa/packaging/pull/138/files#diff-354f30a63fb0907d4ad57269548329e3R14) or [this one](https://github.com/dateutil/dateutil/pull/784/files).

crwilcox:
@di to close

PR #138, opened 2018-10-27
Officially support Python 3.7

colares:
Tested against 3.7 (but there're some pre-existing Warnings)
Added Python 3.7 to classifiers

Issue #135

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: python
 python: 3.6
 
-sudo: false
-
 matrix:
   include:
     - python: 2.7
@@ -16,7 +14,6 @@ matrix:
     - python: 3.7
       env: TOXENV=py37
       dist: xenial
-      sudo: required
     - python: pypy
       env: TOXENV=pypy
     - env: TOXENV=pep8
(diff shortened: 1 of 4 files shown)


jamadden:
`sudo: required` is needed to be able to install 3.7. At any rate, the container-based build infrastructure that you got with `sudo: false` [is going away](https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures). Travis [is asking people to take `sudo: false` out of their configs entirely](https://changelog.travis-ci.com/linux-builds-run-on-vms-by-default-77106), leaving you with the default of `sudo: required`.

.travis.yml
```diff
@@ -13,6 +13,10 @@ matrix:
       env: TOXENV=py35
     - python: 3.6
       env: TOXENV=py36
+    - python: 3.7
+      env: TOXENV=py37
+      dist: xenial
+      sudo: required
```
di: To be absolutely clear: [this line is now the default behavior](https://changelog.travis-ci.com/linux-builds-run-on-vms-by-default-77106) so it is unnecessary. You can also remove the `sudo: false` at the top of this file.

PR #140, opened 2018-10-31
Fix typo in the warning to install readme_renderer[md]

florisla:
Doing "pip install readme_render[md]" (without 'er' after 'render') fails.

I noticed this when doing 'twine check' on a package with a README in MarkDown format.


ssbarnea:
Can we please make a release of this? I drives me mad to read a misleading message.

Issue #142, opened 2018-11-19
No need for a warning about markdown renderers


nedbat:
I have no markdown files in my project, yet when I run "twine check dist/*", I get a warning about "Markdown renderers are not available."  The code looks like it will already issue this warning if markdown is used, so why also issue it on import?  This just adds needless noise to projects that don't use markdown.

di:
This is fixed in https://github.com/pypa/twine/pull/421 but hasn't been released yet.

brainwane:
[Fix is now released!](https://pypi.org/project/twine/1.13.0/)

PR #143, opened 2018-11-19
No need to warn about markdown if it's never used, for #142

nedbat:
Turns out it was super simple.

### DIFF
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -35,7 +35,6 @@ try:
         "CommonMark": cmarkgfm.markdown_to_html,
     }
 except ImportError:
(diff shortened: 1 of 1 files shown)


Julian:
If it matters, I think I agree with @nedbat that even though it's fixed downstream seems unnecessary to warn twice here even if it's fixed downstream too.

nedbat:
I guess in some way it isn't necessary, but only because you know how this file is used in twine.  Looked at in isolation, the file still has an annoying behavior: it has the exact same warning twice: once on import, and then again if the renderer is called.  That seems unnecessary.

nedbat:
Sounds like you don't want this?

PR #164, opened 2020-04-21
Fix regression for image height/width

di:
This fixes a regression introduced in #113 where height/width attributes were removed from images defined in reStructuredText.

More discussion here: https://github.com/pypa/readme_renderer/pull/113#discussion_r206366539

### DIFF
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -27,6 +27,18 @@ class ReadMeHTMLTranslator(HTMLTranslator):
     # Overrides base class not to output `<object>` tag for SVG images.
     object_image_types = {}
 
+    def emptytag(self, node, tagname, suffix="\n", **attributes):
+        """Override this to add back the width/height attributes."""
+        if tagname == "img":
+            if "width" in node:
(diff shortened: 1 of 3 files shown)


matthewfeickert:
Thanks for making this fix PR so quickly, @di. This is really great to see. :)

For the uninitiated like myself who don't know much about how Warehouse works, this will [go into Warehouse](https://github.com/pypa/warehouse/blob/fd7bc0074286296f9952379fdf321167c5f028d6/requirements/main.txt#L607) with release 26.0, correct? Looking at the [release history](https://github.com/pypa/readme_renderer/releases) for `readme_renderer` would this maybe be later in 2020?

di:
Nope, right now: https://pypi.org/project/readme-renderer/26.0/

matthewfeickert:
That's fantastic! :D Thank you so much!

di:
Once the dependency gets pulled into warehouse & merged, old descriptions _should_ get eventually re-rendered, so you shouldn't have to make a new release to see the update (although new releases will definitely get it right away).

PR #192, opened 2021-02-23
Remove py2-specific dependencies and fix tests

mgorny:
1. Fix GitHub actions to actually test against the requested Python version (rather than 2.7 everywhere).
2. Replace external `mock` with py3's `unittest.mock`.
3. Remove dependency on `six`.
4. Use the new `pytest` executable name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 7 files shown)


mgorny:
I dare say the tests are using Python 2.7 instead of the Python they were supposed to test.

mgorny:
I'm going to experiment a little trying to fix it.

mgorny:
Ok, it's ready for review now.

di:
Great catch, thanks!

PR #197, opened 2021-03-01
Remove command-line example from readme

di:
This doesn't currently support Markdown and is confusing to users who don't fully read it.


miketheman (approved):
Removing the untested example is likely best for now.

If we decide that there's more value in command line invocation, we should add testing around that behavior and reintroduce the documentation - possibly with the ideas of #198.

PR #199, opened 2021-04-06
Ignore unknown directive options rather than aborting

terencehonles:
This is the same as my PR https://github.com/github/markup/pull/1452 and it _may_ make sense to incorporate other changes from GitHub's rST processing.

### DIFF
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -15,11 +15,43 @@ from __future__ import absolute_import, division, print_function
 
 import io
 
+from docutils import utils
 from docutils.core import publish_parts
 from docutils.writers.html4css1 import HTMLTranslator, Writer
 from docutils.utils import SystemMessage
 
 from .clean import clean
+
+
+def extract_extension_options(field_list, option_spec):
+    """
+    Overrides `utils.extract_extension_options` and inlines
+    `utils.assemble_option_dict` to make it ignore unknown options passed to
+    directives (i.e. ``:caption:`` for ``.. code-block:``).
+    """
+
+    dropped = set()
+    options = {}
+    for name, value in utils.extract_options(field_list):
+        convertor = option_spec.get(name)
+        if name in options or name in dropped:
+            raise utils.DuplicateOptionError('duplicate option "%s"' % name)
+
+        # silently drop unknown options as long as they are not duplicates
+        if convertor is None:
+            dropped.add(name)
+            continue
+
+        # continue as before
+        try:
+            options[name] = convertor(value)
+        except (ValueError, TypeError) as detail:
+            raise detail.__class__('(option: "%s"; value: %r)\n%s'
+                                   % (name, value, ' '.join(detail.args)))
+    return options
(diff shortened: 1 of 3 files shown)


terencehonles:
I didn't look at the code _too_ closely, but it may make sense to see if these can surfaced as warnings which would still fail with `twine check --strict` but would not prevent the rST from rendering. I had made the change on GitHub since it silently would drop documentation examples and it appears that's the same for PyPI, but since `twine check` and possibly others consume this library the unknown options could possibly still be passed up without blowing up.

terencehonles:
@di not sure if you have any input on this, sorry to bother if you're the wrong person to ask.

readme_renderer/rst.py
```diff
@@ -15,13 +15,45 @@
...
+        # continue as before
+        try:
+            options[name] = convertor(value)
+        except (ValueError, TypeError) as detail:
+            raise detail.__class__('(option: "%s"; value: %r)\n%s'
+                                   % (name, value, ' '.join(detail.args)))
+    return options
+
+
+utils.extract_extension_options = extract_extension_options
```
miketheman, line 54: I can't shake the feeling that this override is "wrong", despite it being completely valid and works. I spent some time trying to find other hooks in docutils to provide an option, a setting, or something to prevent needing to completely override a utility class that is used during rendering, but didn't find anything satisfying.

It works now, but if docutils ever changes the function, would we ever know?

readme_renderer/rst.py
```diff
@@ -15,13 +15,45 @@
...
+    options = {}
+    for name, value in utils.extract_options(field_list):
+        convertor = option_spec.get(name)
+        if name in options or name in dropped:
+            raise utils.DuplicateOptionError('duplicate option "%s"' % name)
+
+        # silently drop unknown options as long as they are not duplicates
+        if convertor is None:
+            dropped.add(name)
+            continue
```
miketheman, line 43: **question:** ‏If I'm reading this correctly, this would drop _any_ unknown options, not only the one surfaced as the `code-block` `caption`, correct?

Does this increase the possibility of errors being introduced for other known options? For example, if I [mistyped the `:header:` option to `:headers:`](https://github.com/pypa/readme_renderer/blame/d821b4466457949aa5f61f951d236fcf5ca170da/tests/fixtures/test_rst_caption.rst#L3-L4), I get no warning/error due to this change, and the headers will be missing from the rendered output.

Would it be better to provide this function a specific list of known options we want to skip and thus keep the scope of the override narrow, at least until [doctuils implements `caption` for `code-block`](https://sourceforge.net/p/docutils/feature-requests/62/)?

miketheman (commented):
Reviewing older PRs, and took some time to discover some history and docutils - good learning on my part!

A few notes inline, using [Conventional Comments](https://conventionalcomments.org/) style.

In general, it needs a rebase against current `main` since we added type hints there's some a conflict.

miketheman:
Related to #160

PR #201, opened 2021-06-05
Allow specific class values only

ewjoachim:
In order to avoid strange things, it would be better to ensure that rendered READMEs only contain the subset of classes that we expect them to contain.

### DIFF
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -41,8 +41,8 @@ ALLOWED_ATTRIBUTES = {
 
     # Custom Additions
     "*": ["id"],
-    "img": ["src", "width", "height", "alt", "align", "class"],
     "span": ["class"],
+    "img": ["src", "width", "height", "alt", "align"],
     "th": ["align"],
     "td": ["align"],
     "h1": ["align"],
@@ -57,6 +57,7 @@ ALLOWED_ATTRIBUTES = {
 # Class is a specific attribute because not only do we want to allow it only
 # on certain tags, but we also want to control possible values.
 ALLOWED_CLASSES = {
+    "img": {"align-left", "align-right", "align-center"},
 }
 
 ALLOWED_STYLES = [
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -50,7 +50,6 @@ ALLOWED_ATTRIBUTES = {
     "h4": ["align"],
     "h5": ["align"],
     "h6": ["align"],
-    "code": ["class"],
     "p": ["align"],
 }
 # Class is a specific attribute because not only do we want to allow it only
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -41,7 +41,6 @@ ALLOWED_ATTRIBUTES = {
 
     # Custom Additions
     "*": ["id"],
-    "hr": ["class"],
     "img": ["src", "width", "height", "alt", "align", "class"],
     "span": ["class"],
     "th": ["align"],
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -41,7 +41,6 @@ ALLOWED_ATTRIBUTES = {
 
     # Custom Additions
     "*": ["id"],
-    "span": ["class"],
     "img": ["src", "width", "height", "alt", "align"],
     "th": ["align"],
     "td": ["align"],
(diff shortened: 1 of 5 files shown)


readme_renderer/clean.py
```diff
@@ -41,9 +41,7 @@
 
     # Custom Additions
     "*": ["id"],
-    "hr": ["class"],
```
ewjoachim, line 44: I wasn't able to find out why `hr` was expected to contain a class. It was added at the same time when `hr` was authorized as a tag itself, by @dstufft in #21 . If you happen to remember what it was about, I'm interested. But I couldn't find a relevant `hr.<somthing>` rule in Warehouse CSS.

readme_renderer/clean.py
```diff
@@ -52,19 +50,43 @@
     "h4": ["align"],
     "h5": ["align"],
     "h6": ["align"],
-    "code": ["class"],
```
ewjoachim, line 55: If we have a `<code class=language-python>print("hello")</code>` in the output of the Markdown rendering, we expressly change it to pygmentize before the bleaching is done, so at bleaching time we should not expect that tag.
https://github.com/pypa/readme_renderer/blob/6a9b282823e033d4344ca519452c49def6d259ec/readme_renderer/markdown.py#L91

readme_renderer/clean.py
```diff
@@ -52,19 +50,43 @@
...
     "p": ["align"],
 }
+# Class is a specific attribute because not only do we want to allow it only
+# on certain tags, but we also want to control possible values.
+ALLOWED_CLASSES = {
+    "img": {"align-left", "align-right", "align-center"},
```
ewjoachim, line 58: If you look at https://github.com/pypa/warehouse/blob/master/warehouse/static/sass/blocks/_project-description.scss#L207-L209, you may find a discrepancy:

Warehouse CSS will align on 3 different conditions:
- `img.align-left` which is the one we target here
- `.figure.align-left` which according to https://github.com/pypa/warehouse/pull/4391#issuecomment-408669826 will be applied on a `div` element, but it's already broken as of today because `div` element doesn't allow `class` without our current bleaching rules, even before this PR. This could be fixed in a subsequent PR
- `object.align-left` but `object` tags are not even allowed!
So authorizing only those classes and only on `img` seems to not be a restriction compared to what we have today.

readme_renderer/clean.py
```diff
@@ -52,19 +50,43 @@
...
+        # Classes for syntax coloring
+        # The original source for this list is
+        # https://github.com/pygments/pygments/blob/cfaa45dcc4103da8cf1700fd0d3e5708d894337b/pygments/token.py
+        # which is a superset from the list in
+        # https://github.com/pypa/warehouse/blob/master/warehouse/static/sass/blocks/_project-description.scss#L256
+        # This means that some classes are unused and it's most probably OK.
+        "bp c c1 ch cm cp cpf cs dl err esc fm g gd ge gh gi go gp gr gs gt "
+        "gu il k kc kd kn kp kr kt l ld m mb mf mh mi mo n na nb nc nd ne nf "
+        "ni nl nn no nt nv nx o ow p py s s1 s2 sa sb sc sd se sh si sr ss sx "
+        "vc vg vi vm w x".split()
```
ewjoachim, line 69: If you wonder, if pygmetize gets new classes, it's not a problem that we won't have them here uless we want to style them in Warehouse, and in this case, we'll have to add them both here and in Warehouse CSS.

tox.ini
```diff
@@ -9,7 +9,7 @@ commands =
 extras = md
 
 [testenv:pep8]
-basepython = python3.6
+basepython = python3
```
ewjoachim, line 12: The CI was launched using [python 3.x](https://github.com/pypa/readme_renderer/blob/main/.github/workflows/ci.yml#L30) but this tox target wanted 3.6 specifically, so it couldn't work. Don't know why it did so far. If you want this in a different PR, let me know.

jboarman:
@ewjoachim 

The [latest `cmarkgfm` release](https://github.com/theacodes/cmarkgfm/releases/tag/0.6.0) makes an adjustment to how code blocks are rendered by removing CSS class assignments and favoring a `lang` attribute instead.

I've not studied the code here, but I did notice your earlier reference to the `code` class and this section of related code:

https://github.com/pypa/readme_renderer/blob/6a9b282823e033d4344ca519452c49def6d259ec/readme_renderer/markdown.py#L91

I'm sharing this in case there is any related concern to what you are working on here in this PR.

ewjoachim:
Indeed! A good thing the dependency is restricted to <0.6.0 in setup.py, so our code won't have to handle both versions!

I guess it's something we'll have to look at when we upgrade but hopefully, when we do this, the impacts will be studied.

ewjoachim:
Oh there's an overlap with #120 !

PR #202, opened 2021-06-05
Add support for GFM task list items

jboarman:
This PR is designed to add support for TODO lists that are provided by [Github Flavored Markdown](https://github.github.com/gfm/#task-list-items-extension-).

The GFM renderer already provides this support.  However, the `clean` operation was filtering out the resulting rendered HTML `input` tags.

To safely add this support, a `disabled` attribute is forced to be added to any rendered HTML `input` tags.

Now, TODO list items are properly rendered from the output of `cmarkgfm.github_flavored_markdown_to_html`.

**Example GFM task list that can now be rendered from a README:**
* [X] Add support for GFM task list items
* [ ] Hide bullets currently rendered by `cmarkgfm` for todo list items

PR #206, opened 2021-07-28
Relax cmarkgfm version requirement

McSinyx:
As confirmed in https://github.com/pypa/readme_renderer/pull/202#issuecomment-887603892, cmarkgfm 0.6.0 is compatible with readme_renderer.  IMHO the minor version shouldn't be pinned so this PR is dropping the upper limit.


McSinyx:
Gentle ping!

McSinyx:
Thank you @benoit-pierre!

Issue #180, opened 2020-11-28
Bug: renderer excludes plain html from readme


HaaLeo:
# Description

on the pypa warehouse readme files which include plain HTML tags are not rendered correctly. The corresponding HTML section is just omitted. This was not the case a few weeks ago. 

# Expected Outcome

HTML tags of a readme are also rendered.

# Actual Outcome

Plain HTML tags are omitted. 

# Example 

The [swarmlib](https://github.com/HaaLeo/swarmlib#swarmlib) project includes a plain paragraph `<p></p>` in its readme:

```
<p align="middle">
  <img src="https://raw.githubusercontent.com/HaaLeo/swarmlib/master/doc/light_mode.png" width="49%" />
  <img src="https://raw.githubusercontent.com/HaaLeo/swarmlib/master/doc/dark_mode.png" width="49%" /> 
</p>
```

But for the latest release the images are not shown correctly on the pypi homepage.
For [`swarmlib==0.12.1`](https://pypi.org/project/swarmlib/0.12.1/) the images are shown correctly. However, for [`swarmlib==0.13.0`](https://pypi.org/project/swarmlib/0.13.0/) no images are rendered in the readme.
Since swarmlib did not change this section of its readme between the last releases I think the `readme_renderer` causes this issue.

di:
Looks like this is due to a change between [`cmarkgfm==0.4.2`](https://pypi.org/project/cmarkgfm/0.4.2/) and [`cmarkgfm==0.5.0`](https://pypi.org/project/cmarkgfm/0.5.0/), probably around it's "safe rendering" features.

I've rolled back the `cmarkgfm` dependency in https://github.com/pypa/warehouse/pull/8877 for now until we can determine exactly what needs updated.

(cc @theacodes)

di:
Looks like it's this breaking change:

> In version 0.4.0 and earlier, the default behavior is unsafe, as described above. In later versions, the default behavior is safe, and to render potentially unsafe HTML pass the `CMARK_OPT_UNSAFE` option.

di:
I made the necessary updates in https://github.com/pypa/readme_renderer/pull/181, will release that tomorrow.

Talendar:
I'm facing the same issue. On the previous releases of my project, all HTML tags were rendered properly. In my new release, however, all HTML tags were simply ignored (along with the content they were "tagging").

di:
[...]

PR #209, opened 2021-09-22
support cmarkgfm>=0.6.0

benoit-pierre:
Closes #206.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,6 @@ setuptools.setup(
     entry_points={
         "distutils.commands": ["check = readme_renderer.integration.distutils:Check"],
     },
-    extras_require={"md": "cmarkgfm>=0.5.0,<0.6.0"},
+    extras_require={"md": "cmarkgfm>=0.5.0"},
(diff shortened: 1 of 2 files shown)


benoit-pierre:
The CI lint job failure is unrelated, see #210 for a fix.

setup.py
```diff
@@ -59,6 +59,6 @@
     entry_points={
         "distutils.commands": ["check = readme_renderer.integration.distutils:Check"],
     },
-    extras_require={"md": "cmarkgfm>=0.5.0,<0.6.0"},
+    extras_require={"md": "cmarkgfm>=0.5.0"},
```
di, line 62:
```suggestion
    extras_require={"md": "cmarkgfm>=0.5.0,<0.7.0"},
```
Since there have been breaking changes between minor versions of `cmarkgfm` in the past (see #180) I'd prefer to keep an upper bound here.

benoit-pierre: OK, done.

di:
Thanks!

PR #225, opened 2022-03-10
chore: add types

### LINKED ISSUE #166: Add type annotations
I've been working on adding type annotations to Twine (https://github.com/pypa/twine/issues/231), and readmer_renderer is one of our imports that's missing type hints/stubs. Is there any interest in adding those, either in the source or to https://github.com/python/typeshed?

See also: https://mypy.readthedocs.io/en/stable/existing_code.html

miketheman:
There's a lot of commits here for a relatively small amount of changes, and it can seem overwhelming at first.

I tried to keep each commit well-scoped, and describe the journey on adding mypy as a testing strategy, correcting the issues surfaced during the initial (non-strict) runs, and then applying stricter rules, until it's finally a fully-typed library.

(I snuck in a pytest warning removal while I was at it.)

Completing this work, and releasing a version should allow us to [remove a line in the warehouse config](https://github.com/pypa/warehouse/blob/83208c50c245cef591c930067dd5fa50b35f27cb/pyproject.toml#L45).

Let me know what y'all think!

Resolves #166

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ requires = ["setuptools>=40.8.0", "wheel", "bleach>=2.1.0", "docutils>=0.13.1", 
 build-backend = "setuptools.build_meta:__legacy__"
 
 [tool.mypy]
+strict = true
 warn_unused_configs = true
 [[tool.mypy.overrides]]
 # These modules do not yet have types available.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,12 @@
 [build-system]
 requires = ["setuptools>=40.8.0", "wheel", "bleach>=2.1.0", "docutils>=0.13.1", "Pygments>=2.5.1"]
 build-backend = "setuptools.build_meta:__legacy__"
+
+[tool.mypy]
+warn_unused_configs = true
+[[tool.mypy.overrides]]
+# These modules do not yet have types available.
+module = [
+    "cmarkgfm.*"
+]
+ignore_missing_imports = true
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -59,9 +59,6 @@ ALLOWED_ATTRIBUTES = {
     "input": ["type", "checked", "disabled"],
 }
 
-ALLOWED_STYLES = [
-]
-
 
 class DisabledCheckboxInputsFilter:
     def __init__(self, source):
@@ -89,19 +86,16 @@ class DisabledCheckboxInputsFilter:
         return getattr(self.source, name)
 
 
-def clean(html, tags=None, attributes=None, styles=None):
+def clean(html, tags=None, attributes=None):
     if tags is None:
         tags = ALLOWED_TAGS
     if attributes is None:
         attributes = ALLOWED_ATTRIBUTES
-    if styles is None:
-        styles = ALLOWED_STYLES
 
     # Clean the output using Bleach
     cleaner = bleach.sanitizer.Cleaner(
         tags=tags,
         attributes=attributes,
-        styles=styles,
         filters=[
             # Bleach Linkify makes it easy to modify links, however, we will
             # not be using it to create additional links.
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -14,6 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import functools
+import typing
 
 import bleach
 import bleach.callbacks
@@ -61,10 +62,10 @@ ALLOWED_ATTRIBUTES = {
 
 
 class DisabledCheckboxInputsFilter:
-    def __init__(self, source):
+    def __init__(self, source: typing.Any) -> None:
         self.source = source
 
-    def __iter__(self):
+    def __iter__(self) -> typing.Iterator[typing.Dict[str, typing.Optional[str]]]:
         for token in self.source:
             if token.get("name") == "input":
                 # only allow disabled checkbox inputs
@@ -82,11 +83,15 @@ class DisabledCheckboxInputsFilter:
             else:
                 yield token
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str) -> typing.Any:
         return getattr(self.source, name)
 
 
-def clean(html, tags=None, attributes=None):
+def clean(
+    html: str,
+    tags: typing.Optional[typing.List[str]] = None,
+    attributes: typing.Optional[typing.Dict[str, typing.List[str]]] = None
+) -> typing.Optional[str]:
     if tags is None:
         tags = ALLOWED_TAGS
     if attributes is None:
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -14,7 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import functools
-import typing
+from typing import Any, Dict, Iterator, List, Optional
 
 import bleach
 import bleach.callbacks
@@ -64,10 +64,10 @@ ALLOWED_ATTRIBUTES = {
 class DisabledCheckboxInputsFilter:
     # The typeshed for bleach (html5lib) filters is incomplete, use `typing.Any`
     # See https://github.com/python/typeshed/blob/505ea726415016e53638c8b584b8fdc9c722cac1/stubs/bleach/bleach/html5lib_shim.pyi#L7-L8 # noqa E501
-    def __init__(self, source: typing.Any) -> None:
+    def __init__(self, source: Any) -> None:
         self.source = source
 
-    def __iter__(self) -> typing.Iterator[typing.Dict[str, typing.Optional[str]]]:
+    def __iter__(self) -> Iterator[Dict[str, Optional[str]]]:
         for token in self.source:
             if token.get("name") == "input":
                 # only allow disabled checkbox inputs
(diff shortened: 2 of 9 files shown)


di (approved):
Thank you! LGTM, I'll give this ~24 hours for others to review before merging.

readme_renderer/rst.py
```diff
@@ -14,20 +14,28 @@
...
 from docutils.core import publish_parts
+from docutils.nodes import colspec, image
 from docutils.writers.html4css1 import HTMLTranslator, Writer
 from docutils.utils import SystemMessage
 
 from .clean import clean
 
 
-class ReadMeHTMLTranslator(HTMLTranslator):
+class ReadMeHTMLTranslator(HTMLTranslator):  # type: ignore
```
bhrutledge, line 27: When using `type: ignore`, it's best practice to indicate why. That could include ignoring a specific error, and/or a reference to a related GitHub issue. See the [mypy docs](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#silencing-type-errors) for details, and an [elaborate example in Twine](https://github.com/pypa/twine/blob/a9e9cd6ecec3c4d4504704ccaba08be487d67ebf/twine/package.py#L127).

Same thing applies to `object_image_types` below.

readme_renderer/clean.py
```diff
@@ -59,15 +60,12 @@
     "input": ["type", "checked", "disabled"],
 }
 
-ALLOWED_STYLES = [
-]
-
 
 class DisabledCheckboxInputsFilter:
-    def __init__(self, source):
+    def __init__(self, source: typing.Any) -> None:
```
bhrutledge, line 65: I'm a little surprised to see this use of `Any`, given the `--strict` configuration. Is there a way to make it a specific type, or a `Union`?

readme_renderer/rst.py
```diff
@@ -95,7 +103,11 @@ def emptytag(self, node, tagname, suffix="\n", **attributes):
 }
 
 
-def render(raw, stream=None, **kwargs):
+def render(
+    raw: str,
+    stream: typing.Optional[typing.Any] = None,
```
bhrutledge, line 108: I _think_ this could be:

```suggestion
    stream: typing.Optional[typing.IO[str]] = None,
```

Via https://mypy.readthedocs.io/en/stable/cheat_sheet.html#miscellaneous

tox.ini
```diff
@@ -1,13 +1,22 @@
...
 extras = md
 
+[testenv:mypy]
+basepython = python3
+deps =
+    mypy
+    types-bleach
+    types-docutils
+    types-Pygments
+commands = mypy --strict readme_renderer
```
bhrutledge, line 18: Glad to see `--strict`. FWIW, I prefer to use an explicit configuration file (ala [twine/mypy.ini](https://github.com/pypa/twine/blob/main/mypy.ini)) and just run `mypy`, because that makes it easier for other tools (e.g. VS Code) to use the same configuration as CI.

pyproject.toml
```diff
@@ -1,3 +1,12 @@
...
 build-backend = "setuptools.build_meta:__legacy__"
+
+[tool.mypy]
+warn_unused_configs = true
+[[tool.mypy.overrides]]
+# These modules do not yet have types available.
+module = [
+    "cmarkgfm.*"
+]
+ignore_missing_imports = true
```
bhrutledge, line 12: Instead of a blanket ignore, I prefer to explicitly ignore packages that are missing imports (e.g. in [twine/mypy.ini](https://github.com/pypa/twine/blob/c506b225b3ab57a40571001825dea0058a55807a/mypy.ini#L21-L42)).

bhrutledge (commented):
Thanks for doing this! At @di's suggestion, I reviewed it based on my experience in Twine. I've got some questions/suggestions about style, but I don't think there's anything that should be considered a blocker.

I confirmed that using this branch allows Twine to remove its [mypy ignore](https://github.com/pypa/twine/blob/c506b225b3ab57a40571001825dea0058a55807a/mypy.ini#L29-L31) for this package.

readme_renderer/markdown.py
```diff
@@ -15,8 +15,9 @@
 
 import re
 import warnings
+from typing import Any, Match, Optional
```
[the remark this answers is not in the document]
miketheman: Heh, good call-out. 
I prefer importing typing and annotating the specifics f I'm there, so as to not have to change the import line if I ever need another type.
It also makes it clearer when reading the code where the annotation comes from.

If there's a pypa preferred style guide, happy to confirm!
This particular inconsistency was me probably trying to beat the pep8 test for line length, and trying to love as few lines of code as possible.

tox.ini
miketheman: Agree! I was using this during dev/test and will move to pyproject.yaml where other configs are.

pyproject.toml
miketheman: This is a difference in syntax between mypy.ini and pyprpject.toml - this flag only affects the modules specified in this overrides section.
I did a similar thing over in warehouse https://github.com/pypa/warehouse/blob/a35c7589d243a71753d9cb1f246989c4f566a95c/pyproject.toml#L18

readme_renderer/clean.py
miketheman: Strict mode doesn't disallow any Any (pun!), rather a subset of all strictness flags.

> Strict mode; enables the following flags: --warn-unused-configs, --disallow-any-generics, --disallow-subclassing-any, --disallow-untyped-calls, --disallow-untyped-defs, --disallow-incomplete-defs, --check-untyped-defs, --disallow-untyped-decorators, --no-implicit-optional, --warn-redundant-casts, --warn-unused-ignores, --warn-return-any, --no-implicit-reexport, --strict-equality

I'll take another look at the potential union for this, and a couple others, but some had to do with not being a subclass with typing.

readme_renderer/rst.py
miketheman: For this line, the parent class is returning Any, and strict mode disallowed that.  I'll clarify the ignore comment

For the other line, I left a lot of detail in the commit message https://github.com/pypa/readme_renderer/pull/225/commits/a0903b16a59c73e3456906d7060b0f79c6a7f802 - I didn't want to pollute the code much with commentary about types.

Is that approach insufficient?

miketheman:
@bhrutledge Thanks for the review!

readme_renderer/clean.py
miketheman: No union type available, upstream types are incomplete, so will add a comment for now.

readme_renderer/rst.py
```diff
@@ -14,20 +14,28 @@
...
 
 
-class ReadMeHTMLTranslator(HTMLTranslator):
+class ReadMeHTMLTranslator(HTMLTranslator):  # type: ignore[misc] # docutils is incomplete, returns `Any` python/typeshed#7256 # noqa E501
 
     # Overrides base class not to output `<object>` tag for SVG images.
-    object_image_types = {}
-
-    def emptytag(self, node, tagname, suffix="\n", **attributes):
+    object_image_types = {}  # type: ignore[var-annotated] # intentnionally empty
```
bhrutledge, line 30: Out of curiosity, I looked at the [docutils source code](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/writers/html4css1/__init__.py#l543). Looks like this could be:

```suggestion
    object_image_types: typing.Dict[str, str] = {}
```

readme_renderer/clean.py
```diff
@@ -85,23 +83,24 @@ def __iter__(self):
...
+    def __getattr__(self, name: str) -> typing.Any:
         return getattr(self.source, name)
 
 
-def clean(html, tags=None, attributes=None, styles=None):
+def clean(
+    html: str,
+    tags: typing.Optional[typing.List[str]] = None,
+    attributes: typing.Optional[typing.Dict[str, typing.List[str]]] = None
+) -> typing.Optional[str]:
```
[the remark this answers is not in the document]
miketheman: Done!

readme_renderer/markdown.py
```diff
@@ -61,7 +66,8 @@ def render(raw, variant="GFM", **kwargs):
     if not renderer:
         return None
 
-    rendered = renderer(raw)
+    # The renderer is a lambda function, and mypy fails lambdas right now.
+    rendered = renderer(raw)  # type: ignore
```
bhrutledge, line 70: Did you try to work around this? I wonder if [`Callable`](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas) and/or [`cast`](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated) would help. Or, maybe `variants` could be updated to use explicit private functions with annotations, instead of lambdas.

miketheman: I did not try to modify the code to make the type annotations work - that felt counter to the effort at hand. mypy struggles with lambdas - see open issue: https://github.com/python/mypy/issues/4226

It'd be awesome if you wanted to take on that as a follow-up once this is merged to remove the ignore!

bhrutledge (approved):
Thanks again for taking my feedback; it was good exercise for me. :)

di:
Released in https://pypi.org/project/readme-renderer/34.0/.

PR #230, opened 2022-03-12
Fix renderer ignore

bhrutledge:
Follow-up to #225.

Based on: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas

mypy can infer lambda types; it seems we just need to annotate the lookup dict.

The `cast` isn't necessary to pass mypy, but offers a little more clarity.

FYI @miketheman

### DIFF
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -67,7 +67,7 @@ def render(
         return None
 
     # The renderer is a lambda function, and mypy fails lambdas right now.
-    rendered = renderer(raw)  # type: ignore
+    rendered = renderer(raw)  # type: ignore[no-untyped-call]
 
     if not rendered:
         return None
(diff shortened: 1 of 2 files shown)


readme_renderer/markdown.py
```diff
@@ -33,13 +33,13 @@
 try:
     import cmarkgfm
     from cmarkgfm.cmark import Options as cmarkgfmOptions
-    variants = {
-        "GFM": lambda raw: cmarkgfm.github_flavored_markdown_to_html(
+    variants: Dict[str, Callable[[str], str]] = {
+        "GFM": lambda raw: cast(str, cmarkgfm.github_flavored_markdown_to_html(
```
miketheman, line 37: Nifty trick, I'll have to remember it for the future. I've confirmed this was included in CPython since 3.5, so we're definitely safe there.

miketheman (commented):
Nice stuff, thanks for doing it!

miketheman (approved):

di (approved):

PR #231, opened 2022-03-13
fix: emit a warning when no content is rendered

### LINKED ISSUE #149: RST cannot contain header only - cause silent crash
If the source README.rst contains header only, e.g.:
```rst
Header
======
```
The `twine check` command claims the markup is invalid for no reason:
```
twine check 'dist/*'
Checking distribution dist/foo-0.0.1.tar.gz: Failed
The project's long_description has invalid markup which will not be rendered on PyPI. The following syntax errors were detected:
```
Without listing any errors. Thats the end of `stdout/stderr`.

If I change the `README.rst` by adding just single letter at end, e.g.:
```rst
Header
======
a
```
or clear it contents everything is ok:
```
twine check 'dist/*'
Checking distribution dist/grot-0.0.1.tar.gz: Passed
```
I debugged it a bit. And got to the line 99 in `site-packages/readme_renderer/rst.py`:
`rendered = parts.get("docinfo", "") + parts.get("fragment", "")`
and the `rendered` string is empty. (both `docinfo` and `fragment` contain empty strings).
It corrupts the later `if`:
```py
    if rendered:
        return clean(rendered)
    else:
        return None
```
...causing `render` function return `None`, which is interpreted by twine.check as error, but no warnings/errors are collected by used  `stream`.

miketheman:
Fixes #149

### DIFF
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -14,7 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import io
-from typing import Any, Dict, Optional, TextIO, Union, TYPE_CHECKING
+from typing import Any, Dict, IO, Optional, Union
 
 from docutils.core import publish_parts
 from docutils.nodes import colspec, image
@@ -22,10 +22,6 @@ from docutils.writers.html4css1 import HTMLTranslator, Writer
 from docutils.utils import SystemMessage
 
 from .clean import clean
-
-if TYPE_CHECKING:
-    # Prevent a circular import
-    from .integration.distutils import _WarningStream
 
 
 class ReadMeHTMLTranslator(HTMLTranslator):  # type: ignore[misc] # docutils is incomplete, returns `Any` python/typeshed#7256 # noqa E501
@@ -109,7 +105,7 @@ SETTINGS = {
 
 def render(
     raw: str,
-    stream: Optional[Union[io.StringIO, TextIO, "_WarningStream"]] = None,
+    stream: Optional[IO[str]] = None,
     **kwargs: Any
 ) -> Optional[str]:
     if stream is None:
@@ -134,16 +130,6 @@ def render(
         return clean(rendered)
     else:
         # If the warnings stream is empty, docutils had none, so add ours.
-
-        # We have to currently handle the legacy disutils integration check
-        # command, as it implements its own warnings stream structure.
-        # Once that code is removed, we can remove this conditional.
-        if hasattr(stream, "output"):
-            if not stream.output.getvalue():  # type: ignore[union-attr]
-                stream.write("No content rendered from RST source.")
-        elif isinstance(stream, io.StringIO) and not stream.getvalue():
-            stream.write("No content rendered from RST source.")
-        elif isinstance(stream, TextIO) and not stream.readlines():
-            # This is `sys.stderr` from using readme_renderer CLI
+        if not stream.tell():
             stream.write("No content rendered from RST source.")
         return None
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -14,7 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import io
-from typing import Any, Dict, IO, Optional, Union
+from typing import Any, Dict, Optional, TextIO, Union, TYPE_CHECKING
 
 from docutils.core import publish_parts
 from docutils.nodes import colspec, image
@@ -22,6 +22,10 @@ from docutils.writers.html4css1 import HTMLTranslator, Writer
 from docutils.utils import SystemMessage
 
 from .clean import clean
+
+if TYPE_CHECKING:
+    # Prevent a circular import
+    from .integration.distutils import _WarningStream
 
 
 class ReadMeHTMLTranslator(HTMLTranslator):  # type: ignore[misc] # docutils is incomplete, returns `Any` python/typeshed#7256 # noqa E501
@@ -105,7 +109,7 @@ SETTINGS = {
 
 def render(
     raw: str,
-    stream: Optional[IO[str]] = None,
+    stream: Optional[Union[io.StringIO, TextIO, "_WarningStream"]] = None,
     **kwargs: Any
 ) -> Optional[str]:
     if stream is None:
@@ -129,5 +133,17 @@ def render(
     if rendered:
         return clean(rendered)
     else:
-        stream.write("no output rendered")
+        # If the warnings stream is empty, docutils had none, so add ours.
+
+        # We have to currently handle the legacy disutils integration check
+        # command, as it implements its own warnings stream structure.
+        # Once that code is removed, we can remove this conditional.
+        if hasattr(stream, "output"):
+            if not stream.output.getvalue():  # type: ignore[union-attr]
+                stream.write("No content rendered from RST source.")
+        elif isinstance(stream, io.StringIO) and not stream.getvalue():
+            stream.write("No content rendered from RST source.")
+        elif isinstance(stream, TextIO) and not stream.readlines():
+            # This is `sys.stderr` from using readme_renderer CLI
+            stream.write("No content rendered from RST source.")
         return None
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -129,4 +129,5 @@ def render(
     if rendered:
         return clean(rendered)
     else:
+        stream.write("no output rendered")
         return None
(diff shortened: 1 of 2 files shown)


miketheman:
cc @bhrutledge - the bug surfaced via a `twine check` command - see original issue.
This library hadn't previously emitted any specific warnings, rather passing through any received from `docutils`, so this is the first that I can tell where we're emitting our own to the stream.

bhrutledge:
@miketheman Thanks for sharing, and fixing! I'd like to take a look at how this displays in Twine, and see if there could be a more helpful message. I should be able to do that later today.

readme_renderer/rst.py
```diff
@@ -129,4 +129,5 @@ def render(
     if rendered:
         return clean(rendered)
     else:
+        stream.write("no output rendered")
```
bhrutledge, line 132: I think this could use some more work. Given a `README.rst` like:

```rst
Header
======
```

The latest (unreleased) version of Twine shows:

Screen Shot 2022-03-13 at 9 17 31 PM

First, it'd be nice if the `no output rendered` message were more helpful, e.g. with a suggested resolution.

Furthermore, this new output also shows up for RST syntax errors. With a spurious header line:

```rst
Header
======

======
```

The output is:

Screen Shot 2022-03-13 at 9 17 02 PM

Compared to `readme_renderer==34.0`:

Screen Shot 2022-03-13 at 9 15 58 PM

---

Also, I wonder if the scenario described in #149 (automatically removed header) could/should be treated the same way as an empty `README.rst`, i.e., passing with an empty description:

Screen Shot 2022-03-13 at 9 15 47 PM

bhrutledge (changes requested):

readme_renderer/rst.py
miketheman: @bhrutledge Thanks for testing this branch - these examples are very helpful!

Considering that the condition expressed in the added test case is [intentional behavior](https://github.com/pypa/readme_renderer/issues/149#issuecomment-1067261247), would it make sense to not apply this change, and rather change how twine behaves, rather than this library?
Maybe around https://github.com/pypa/twine/blob/a0ba32dcd0ea5af5de7b88d6645b7743bf003760/twine/commands/check.py#L102-L103 - check if there are warnings on the stream as well? This would then preserve the desired behavior of `readme_renderer` to emit the warnings from the docutils dependency it uses.

Then the second example with the spurious header line would still work as it's supposed to.

bhrutledge: Twine could check the warning stream, but if I follow this code, so could readme_renderer, and add a helpful warning if there's no content. I think I have a preference for the latter, to make this transparent to Twine, and useful in other projects that use this package. What do you think?

miketheman: Fair enough - I figured if the original report was expressing dissatisfaction with their `twine check` results, maybe twine would answer their need. 😉 

I'll give it a shot - checking the stream for *any warnings from docutils* first, and if there are, leave them be, but if there are none, then add our warning.

Do you have any suggestions for a better warning message?

readme_renderer/rst.py
```diff
@@ -129,4 +133,17 @@ def render(
...
         return clean(rendered)
     else:
+        # If the warnings stream is empty, docutils had none, so add ours.
+
+        # We have to currently handle the legacy disutils integration check
+        # command, as it implements its own warnings stream structure.
+        # Once that code is removed, we can remove this conditional.
+        if hasattr(stream, "output"):
+            if not stream.output.getvalue():  # type: ignore[union-attr]
+                stream.write("No content rendered from RST source.")
```
miketheman, line 143: @di The `check` command was deprecated back in https://github.com/pypa/readme_renderer/pull/122 - are we fine with removing that code? It'd make this code a little cleaner if it didn't have to consider that command.

tests/test_rst.py
```diff
@@ -53,3 +53,33 @@ def test_rst_raw():
 """, stream=warnings) is None
 
     assert '"raw" directive disabled' in warnings.getvalue()
+
+
+def test_rst_empty_file():
+    warnings = io.StringIO()
+    assert render("", stream=warnings) is None
```
miketheman, line 60: @bhrutledge I added an example for an empty RST file which should now emit a warning.

readme_renderer/rst.py
miketheman: @bhrutledge Thanks for digging in - super helpful!
I've pushed another commit (after rebase of the original one on the current `main`) - which implements the clearer message you suggested, and added some smarter detection of whether or not to emit a warning.

Glad that we exposed an interesting bug on `twine` as well - but considering that with this change `readme_renderer` will emit a warning for what it considers "empty" (after removing the first Header), does the `UNKNOWN` check still make sense over there? 🤔

bhrutledge: > with this change `readme_renderer` will emit a warning for what it considers "empty" (after removing the first Header), does the `UNKNOWN` check still make sense over there? 🤔

I want to look into this a little more, but I think so, because otherwise readme_renderer will see `UNKNOWN\n\n`, and return `<p>UNKNOWN</p\n`. What's not clear to me at the moment is how that's rendered on PyPI, and the whole point of `twine check` is to avoid surprises in PyPI rendering.

readme_renderer/rst.py
bhrutledge: The answer here might make this moot, but looking at the code for `Check`:

https://github.com/pypa/readme_renderer/blob/361f7fc598c9333f6ba3755e9f80e171d0bdce5a/readme_renderer/integration/distutils.py#L97-L105

It seems like it could be reworked to remove `_WarningStream`, and pass in `io.StringIO` instead. That would simplify this conditional, and remove the circular import.

readme_renderer/rst.py
```diff
@@ -129,4 +133,17 @@ def render(
...
+        elif isinstance(stream, io.StringIO) and not stream.getvalue():
+            stream.write("No content rendered from RST source.")
+        elif isinstance(stream, TextIO) and not stream.readlines():
+            # This is `sys.stderr` from using readme_renderer CLI
+            stream.write("No content rendered from RST source.")
```
bhrutledge, line 148: I'm a little wary of consuming the content on the `TextIO`. Following my suggestion re: `Check`, I think this could be simplified to:

```python
if not stream.read():
    stream.write("No content rendered from RST source.")
stream.seek(0)
```

However it works out, it'd be nice if the `"No content..."` literal was only defined once.

readme_renderer/rst.py
bhrutledge: Although it looks like Twine would need to make a similar change (which I’m happy to do):

https://github.com/pypa/twine/blob/a6dd69c79f7b5abfb79022092a5d3776a499e31b/twine/commands/check.py#L51

miketheman: Instead of reworking this deprecated code, I've prepared #233 to remove it, and if that gets merged, I can simplify this.

bhrutledge: Nice to see it removed, but to be clear, not checking for `stream.output` means Twine will also need to replace the `_WarningStream` class. I'm happy to do that, but it'll be this weekend at the earliest. It'd be nice if we held off on releasing `readme_renderer` until that's done, to avoid breaking current workflows.

miketheman: @bhrutledge no worries, makes sense! This code isn't ready to merge yet - I'm updating it now to account for the removal, and will convert to Draft to make sure someone doesn't accidentally merge until twine is updated.

bhrutledge: Twine 4.0.0 is released, which should support this change.

miketheman:
@bhrutledge I think this is now ready for re-review and release, considering that twine 4.0.0 is out with the changes to the stream.

bhrutledge:
@miketheman Thanks for the nudge. It's been on my list, blocked by "real life". 😉  Hopefully this weekend.

readme_renderer/rst.py
```diff
@@ -129,4 +129,7 @@ def render(
     if rendered:
         return clean(rendered)
     else:
+        # If the warnings stream is empty, docutils had none, so add ours.
+        if not stream.tell():
```
bhrutledge, line 133: Clever!

bhrutledge (approved):
I think this is good! Here's the result from an empty README and a README with only a header:

```
% twine check dist/*.gz 
Checking dist/example-rst-empty-0.0.5.tar.gz: PASSED with warnings
WARNING  `long_description` missing.                                                                                                                                                             
Checking dist/example-rst-header-0.0.5.tar.gz: FAILED
ERROR    `long_description` has syntax errors in markup and would not be rendered on PyPI.                                                                                                       
         No content rendered from RST source.  
```

miketheman:
Thanks! @di all yours now.

di (approved):

PR #233, opened 2022-03-23
chore: remove legacy check command and distutils

miketheman:
The `check` command was ported to `twine check` in 2018.
Refs: #122

Python 3.10 and 3.11 have marked `distutils` as deprecated.
Refs: https://peps.python.org/pep-0632/#specification


miketheman:
Uncovered during work on #231 and would simplify that implementation once this is removed.

di:
Thanks!

PR #236, opened 2022-04-19
chore: drop support for python 3.6

miketheman:
It's been EOL since 23 Dec 2021 and is now no longer passing tests.

If someone does still use this library with python 3.6, they will need
to pin the `bleach` dependency to 4.1.0 and `readme_renderer` 34.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


miketheman:
@di I think this should be ready for review, but there's a branch protection rule on `main` expecting "Test - 3.6" and "Test - pypy3"  to be reported before it can be merged. 3.6 was removed and `pypy3` was renamed to `pypy-3.7`.

The tests won't pass, since the change for the tests are running on `bleach` 5.

PR #238, opened 2022-04-19
feat(dev): enable gitpod development

miketheman:
Use the initalization phase to install all of the Python versions
and run the tests on startup, making subsequent test runs faster.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,23 @@
+# See https://www.gitpod.io/docs/references/gitpod-yml for full reference
+
+tasks:
+  - name: Setup Development and run Tests
+
+    init: |
+      # Upgrade pyenv itself
+      pyenv update
+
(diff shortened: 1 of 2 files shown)


miketheman:
I don't think we have GitPod enabled on the `pypa` org yet, but this could be a starter on a simpler repo to demonstrate the desired behaviors.

Test this out with a URL like: https://gitpod.io/#https://github.com/pypa/readme_renderer/pull/238 (use your own personal GitHub auth).

Without [Prebuilds](https://www.gitpod.io/docs/prebuilds) enabled, it'll take some time to run all the `init` commands, but eventually completes and runs `tox` to prove a "clean slate" before the developer makes any changes.

We could also try to create our own custom `Dockerfile` instead, but that seemed like a heavier lift, since we'd have to push the image to DockerHub or another registry, etc.

Looking forward to your thoughts!

PR #245, opened 2022-07-03
fix: enable `colspan` and `rowspan` attributes

### LINKED ISSUE #182: RST tables multirow cells rendering
As I understand this is the lib that powers PyPI REDME rendering so I post this question here.

## The problem

PyPI renders multirow cells in a way that it essentially ignores them which messes up the following rows. I have first encounterd this problem when trying to publish package to PyPI: https://pypi.org/project/ssh-utilities/ .

## Minimal reproducible example

```rst
+---------------+---------------+
| title1        | title2        |
+===============+===============+
| col1          | col2          |
+---------------+---------------+
| mutirow       | cell1         |
|               +---------------+
|               | cell2         |
+---------------+---------------+
| singlerow     | cell3         |
+---------------+---------------+
```

Render this rst code with `readme-renderer`

```bash
python -m readme_renderer README.rst -o README.html
```

You will get this:
example

In the picture you can see the **cell2** is incorrectly possitioned in the first column.

As far as I know RST is not supposed to render like this. So is this a know bug? Does any workaround exist? or is this on purpose?

miketheman:
Previously exlcuded directives made for incorrect rendering of rst table
cells spanning more than a single column or row.

Fixes #182

### DIFF
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -46,7 +46,7 @@ ALLOWED_ATTRIBUTES = {
     "img": ["src", "width", "height", "alt", "align", "class"],
     "span": ["class"],
     "th": ["align"],
-    "td": ["align"],
+    "td": ["align", "colspan", "rowspan"],
     "div": ["align", "class"],
     "h1": ["align"],
     "h2": ["align"],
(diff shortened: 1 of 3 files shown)


miketheman:
While debugging this, I created a jsfiddle that helped demonstrate the visual differences from the code before and after:
https://jsfiddle.net/miketheman/mpzavs5h/ (see results panel)

di (approved):

PR #248, opened 2022-07-19
Release 36.0

miketheman:


miketheman:
I've tested the math behavior sufficiently that the output HTML does indeed render the necessary HTML for MathJax to pick up and render correctly.

Here's an example screenshot:
Screen Shot 2022-08-06 at 10 01 37 AM

di (approved):

PR #251, opened 2022-08-07
Multiple pyproject.toml fixes

mgorny:
1. Remove the redundant `wheel` requirement
2. Remove the incorrect build-time dependencies from `requires`
3. Use the correct setuptools backend


miketheman (approved):
I've confirmed that the resulting `requires.txt` file in the package remains the same and contains the correct requirements.
This should reduce the amount of dependencies that are installed for **builds** - but tests will still install the requirements per the test suites.

Thanks!

mgorny:
Thanks!

PR #271, opened 2023-01-02
Add CLI option to render package README.

tillahoffmann:
This PR adds an option `--package` to the CLI such that the `README` of a package can be rendered as it would on `pypi.org`. Using this option, package authors can preview `README`s in addition to checking their validity using `twine check`. Rendering the `README` based on the package metadata is important because a number of packages strip out roles and directives that are not supported by `pypi.org` in their `setup.py`. Publishing to `test.pypi.org`, as discussed in pypa/twine#875, is an alternative, but having a local preview would be desirable (at least for myself). This PR also adds tests for the existing CLI.

### DIFF
--- a/readme_renderer/__main__.py
+++ b/readme_renderer/__main__.py
@@ -1,29 +1,60 @@
 import argparse
 import email
-from readme_renderer.rst import render
+from readme_renderer.markdown import render as render_md
+from readme_renderer.rst import render as render_rst
+from readme_renderer.txt import render as render_txt
+import pathlib
 from pkg_resources import get_distribution
 import sys
 
 
 def __main__(args=None):
     parser = argparse.ArgumentParser(
-        description="Renders a .rst README to HTML",
+        description="Renders a .md, .rst, or .txt README to HTML",
     )
     parser.add_argument("-p", "--package", help="Get README from package metadata",
                         action="store_true")
+    parser.add_argument("-f", "--format", choices=["md", "rst", "txt"],
+                        help="README format (inferred from input file name or package)")
     parser.add_argument('input', help="Input README file or package name")
     parser.add_argument('-o', '--output', help="Output file (default: stdout)",
                         type=argparse.FileType('w'), default='-')
     args = parser.parse_args(args)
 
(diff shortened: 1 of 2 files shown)


miketheman:
@tillahoffmann Thanks! We removed the documented example of using the CLI in https://github.com/pypa/readme_renderer/pull/197 - since the interface doesn't currently render Markdown. Before introducing a CLI that we'd expect to support, would you consider adding that behavior?

Please also update your branch from `main` as it includes some test fixes.

readme_renderer/__main__.py
```diff
@@ -1,19 +1,65 @@
...
+            content_type = message.get("Description-Content-Type", "text/x-rst")
+            if content_type == "text/x-rst":
+                content_format = "rst"
+            elif content_type == "text/markdown":
+                content_format = "md"
+            elif content_type == "text/plain":
+                content_format = "txt"
+            else:
+                raise ValueError(f"invalid content type {content_type} for package "
+                                 "`long_description`")
```
tillahoffmann, line 42: This part isn't particularly well tested because we'd have to install a bunch of dummy packages for the tests to run. Or we could mock out `get_distribution` to return the right data.

readme_renderer/__main__.py
```diff
@@ -1,19 +1,65 @@
...
+from readme_renderer.markdown import render as render_md
+from readme_renderer.rst import render as render_rst
+from readme_renderer.txt import render as render_txt
+import pathlib
+from pkg_resources import get_distribution
 import sys
 
 
-if __name__ == '__main__':
+def __main__(args=None):  # noqa: N807
```
miketheman, line 11: **suggestion:** ‏instead of disabling the linter here, how about rename the function to be `main()` instead?

tillahoffmann:
Hm, interesting, any idea why pypy might fail while the others succeed? I can also mock out writing to the file rather than using a temporary directory.

miketheman (approved):
Thanks for working through this!

tillahoffmann:
> Thanks for working through this!

Thanks for the speedy review!

PR #272, opened 2023-01-04
Adapt tests to pygments 2.14.0

danigm:
This patch update the test fixtures to match the output of pygments 2.14.0


miketheman:
Thanks @danigm ! I'm curious - how did this surface, and do you happen to know what changed in Pygments that caused this change in the output?

danigm:
> Thanks @danigm ! I'm curious - how did this surface, and do you happen to know what changed in Pygments that caused this change in the output?

I found the problem after fixing a similar issue in pytest: https://github.com/pytest-dev/pytest/issues/10630

What I've done to change the output is compare what the test output is expecting and modifying the fixtures until all tests are correct, but you can check the [pygments commit](https://github.com/pygments/pygments/commit/147b22face65617514ccfa8512c6b097b07cad4c) and see all the changes in examplefiles, that are very similar to this one.

di:
Thanks!

PR #276, opened 2023-04-26
Update `release.yml` to use Trusted Publishing

di:
https://docs.pypi.org/trusted-publishers/


miketheman (approved):
I've noticed I'm not in an org/group of admins for this package on pypi yet

di:
> I've noticed I'm not in an org/group of admins for this package on pypi yet

Invite sent!

PR #282, opened 2023-06-09
chore: drop support for python 3.7

miketheman:
Python 3.7 is no longer supported, so we can drop support here as well. Using the `python_requires` flag in the package metadata should guide installers on 3.7 to use the oldest published version (37.3).


miketheman:
Needs a repo admin to remove the Required status for `Test - 3.7`

di (approved):

PR #288, opened 2023-07-07
test: test our own readme

miketheman:
Verify that it doesn't get broken by our own library.

Note: This file is also covered by the `test_cli_package` test case, but this makes it super explicit.


di (approved):
Glad this passes but I wasn't expecting it to!

miketheman:
> Glad this passes but I wasn't expecting it to!

Same, mildly surprised. Makes me think the bug is in warehouse, not here.

Issue #284, opened 2023-06-18
Convert package to PEP621 pyproject.toml


miketheman:
https://peps.python.org/pep-0621/
https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata

Some considerations need to be taken for  current setup.py dynamic attributes from `__about__.py`:

https://github.com/pypa/readme_renderer/blob/c6f1eabd997f4f1f2cb6edf727790c83bba3911c/setup.py#L21-L23

https://github.com/pypa/readme_renderer/blob/c6f1eabd997f4f1f2cb6edf727790c83bba3911c/setup.py#L30-L37

Some inspiration could be drawn from here: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

Other values might be better served as static metadata.

di:
Resolved by #293.

PR #293, opened 2023-08-23
PEP 621 added

apurvakhatri:
Issue number of the reported bug or feature request: #284 

Describe your changes
Read PEP 621 Documentation and migrated metadata from setup.py to pyproject.toml. 
project table contains the static fields whereas version and classifiers are added dynamic fields as per PEP 621.

Testing performed
python3 setup.py bdist and twine check dist/ ran successfully. 
OS: MacOS 13.3.1

Additional context
.toml does not support reusing of variables (https://github.com/toml-lang/toml/issues/829) therefore it was difficult to migrate __copyright__, which uses __author__.


setup.py
```diff
@@ -27,34 +27,8 @@
...
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
-        "Programming Language :: Python :: 3.11",
-        "Programming Language :: Python :: Implementation :: CPython",
-        "Programming Language :: Python :: Implementation :: PyPy",
-        "Typing :: Typed",
-    ],
+    long_description=long_description,
     install_requires=["bleach>=2.1.0", "docutils>=0.13.1", "Pygments>=2.5.1"],
```
di, line 32: We need to move `install_requires` and `extras_require` into `pyproject.toml` as well: https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dependencies-optional-dependencies

setup.py
```diff
@@ -27,37 +27,9 @@
...
-        "Programming Language :: Python :: Implementation :: PyPy",
-        "Typing :: Typed",
-    ],
-    install_requires=["bleach>=2.1.0", "docutils>=0.13.1", "Pygments>=2.5.1"],
+    long_description=long_description,
     include_package_data=True,
-    extras_require={"md": "cmarkgfm>=0.8.0"},
     packages=setuptools.find_packages(exclude=["tests", "tests.*"]),
     python_requires=">=3.8",
 )
```
di, line 35: I think we want to get to the point where everything has been moved to `pyproject.toml` and the `setup.py` file doesn't exist at all -- I don't think we can mix metadata between the two files like this.

apurvakhatri: I removed the dependency of setup.py completely. Also removed __about__.py since only the classifiers are dynamic. 
I could not figure out how to add __copyright__ into pyproject.toml. I can integrate it, but could not find variables conveying same information. 

Testing: 
The distribution was build using PyPa build.
`python3 -m build
`
Further, the dist, PASSED twine check.

pyproject.toml
```diff
@@ -1,3 +1,34 @@
...
+version = "41.0"
+description = '''readme_renderer is a library for rendering "readme" ' "descriptions for Warehouse'''
+authors = [
+  {email = "admin@mail.pypi.org"},
+  {name = "The Python Packaging Authority"}
+]
+readme = "README.rst"
+license = {file = "LICENSE"}
+dependencies = ["bleach>=2.1.0", "docutils>=0.13.1", "Pygments>=2.5.1"]
+dynamic = ["classifiers"]
```
miketheman, line 12: **question:** ‏Can we move the classifiers inline to pyproject.toml and remove the `dynamic` nature key here as well as the `[tool.setuptools.dynamic]` section? This reduces an extra file in the root of the project, as well as one setuptools-specific behavior.

See https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#classifiers for more

setup.py
miketheman, line 1: **question:** ‏with setup.py now gone, do we still need [`MANIFEST.in`](https://github.com/pypa/readme_renderer/blob/9c0f7d8d141abff4fcae4b9273ac385c6bbe9cf3/MANIFEST.in)? Are all the directives specified reflected in the new packages? With the new build, do files like `py.typed` end up in the distribution package?

readme_renderer/__about__.py
```diff
@@ -1,38 +0,0 @@
...
-)
-__uri__ = "https://github.com/pypa/readme_renderer"
-
-__version__ = "41.0"
-
-__author__ = "The Python Packaging Authority"
-__email__ = "admin@mail.pypi.org"
-
-__license__ = "Apache License, Version 2.0"
-__copyright__ = "Copyright 2014 %s" % __author__
```
miketheman, line 38: **question:** ‏do we completely lose this ? I realize it's an invalid metadata key - but I don't think we've recorded the copyright notice anywhere else. Maybe that should at least be reflected in the README.

pyproject.toml
```diff
@@ -1,3 +1,34 @@
+[project]
+name = "readme_renderer"
+version = "41.0"
+description = '''readme_renderer is a library for rendering "readme" ' "descriptions for Warehouse'''
+authors = [
+  {email = "admin@mail.pypi.org"},
+  {name = "The Python Packaging Authority"}
```
miketheman, line 7: **suggestion:** ‏This is specifying two entries in the `authors` value, instead of one being a name, email pair.
See the examples show in https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#authors-maintainers for the exact syntax.

pyproject.toml
```diff
@@ -1,3 +1,34 @@
...
+readme = "README.rst"
+license = {file = "LICENSE"}
+dependencies = ["bleach>=2.1.0", "docutils>=0.13.1", "Pygments>=2.5.1"]
+dynamic = ["classifiers"]
+
+[project.optional-dependencies]
+md = ["cmarkgfm>=0.8.0"]
+
+[tool.poetry.dependencies]
+python = ">=3.8"
```
miketheman, line 18: **question:** ‏are we also using poetry tool, or is this a copy/paste from another example?
See https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement for setuptools syntax

miketheman (changes requested):
Coming along nicely, thanks for working through this! I've added some notes inline - let me know if anything is unclear!

setup.py
apurvakhatri: Test: Removed MANIFEST.in 
Result: test/*, py.typed, README.rst, CHANGES.rst and tox.ini are included in the dist.

pyproject.toml
```diff
@@ -1,3 +1,47 @@
+[project]
+name = "readme_renderer"
+version = "41.0"
+description = '''readme_renderer is a library for rendering "readme" ' "descriptions for Warehouse'''
```
miketheman, line 4: **suggestion:** ‏this line looks to be concatenated two parts of a string - with extra quotes in the middle.  I am pretty sure we can remove some of the extras in the middle and wrap the entire single line with a single quote instead.

setup.py
miketheman: See the test failure here: https://github.com/pypa/readme_renderer/actions/runs/5981340430/job/16230249111?pr=293
I think we might have to add some specific includes to make them be part of of the package.

miketheman (changes requested):

pyproject.toml
```diff
@@ -1,3 +1,47 @@
...
+md = ["cmarkgfm>=0.8.0"]
+
+[tool.setuptools]
+include-package-data = true
+
+[tool.setuptools.packages.find]
+exclude = ["tests", "tests.*"]
+
+[project.urls]
+urls = "https://github.com/pypa/readme_renderer"
```
miketheman, line 43:
```suggestion
Home-page = "https://github.com/pypa/readme_renderer"
```
Per https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#urls

pyproject.toml
```diff
@@ -1,3 +1,47 @@
+[project]
+name = "readme_renderer"
+version = "41.0"
+description = "readme_renderer is a library for rendering readme descriptions for Warehouse"
+authors = [
+  {name = "The Python Packaging Authority", email = "admin@mail.pypi.org"}
+]
+readme = "README.rst"
+license = {file = "LICENSE"}
```
miketheman, line 9: This directive changes the metadata to **embed** the license text into the metadata, previously this would be a string of the license type, not the entire file.
This change would preserve the value from `__about__.py`:
```suggestion
license = {text = "Apache License, Version 2.0"}
```

di (approved):
Thanks!

PR #295, opened 2023-09-01
refactor: replace bleach with nh3 (ammonia) 

miketheman:
Refs: https://nh3.readthedocs.io/

Bleach has been deprecated.

This refactor includes"
- a non-breaking refactor of lists to sets to support both bleach and nh3
- addition of the `pytest-icdiff` test helper to find the one-character changes
- removal of the bleach code and filters
- updates to the test suite

Notable changes in the tests:

- **all** `<a>` tags now get `rel="nofollow"` added
- less usage of escaped characters like `&quot;` for `"`


miketheman:
Expecting failure for `checked` boolean attributes - [see this issue](https://github.com/rust-ammonia/ammonia/issues/186) for more.

miketheman:
We _could_ update the tests here as well to allow `checked` to become `checked=""` in the meantime, if we were worried about continued use of `bleach`.

di:
> We could update the tests here as well to allow checked to become checked="" in the meantime, if we were worried about continued use of bleach.

I think that's probably fine. Since they're equivalent, I don't think it really matters which one ends up being used.

di (approved):

PR #307, opened 2024-06-20
Exclude .gitpod.yml by default with check-manifest

shenxianpeng:
since https://github.com/mgedmin/check-manifest/pull/155 has been released, it's time to remove .gitpod.yml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,9 +47,7 @@ Home-page = "https://github.com/pypa/readme_renderer"
 requires = ["setuptools>=40.8.0"]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -50,9 +50,7 @@ Home-page = "https://github.com/pypa/readme_renderer"
 requires = ["setuptools>=40.8.0"]
 build-backend = "setuptools.build_meta"
 
-# TODO: Remove when https://github.com/mgedmin/check-manifest/pull/155 released
 [tool.check-manifest]
```
miketheman, line 53: I think now that there's no need for the rule, is there any need for a config section?
```suggestion
```

miketheman (approved):

PR #312, opened 2024-06-29
Lint specific directories

kurtmckee:
My workflow includes creating a virtual environment in `.venv/` on both Linux and Windows and installing the package in editable mode. This causes `flake8 .` to churn through everything in the virtual environment and throw hundreds of errors.

This PR modifies the `flake8 .` lint command to target the project's code directories. It also configures the tox environment to skip building/installing the package, since it's unnecessary.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,10 +28,11 @@ commands = mypy readme_renderer
 
 [testenv:pep8]
 basepython = python3
+skip_install = true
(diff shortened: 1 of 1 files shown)


kurtmckee:
_Rebased on `main`_

miketheman (approved):

PR #308, opened 2024-06-27
Build a wheel once, for all test environments

kurtmckee:
The test suite currently builds a `.tar.gz` file, but when it's installed in each tox environment, pip converts the `.tar.gz` file to a wheel and _then_ installs it.

This PR eliminates the intermediary `.tar.gz` file and just builds the wheel once for all environments.

----

For verification, you can run `tox -vvvvve py312` before and after.

### Package installation step (before)

```
py312: 874 W install_package> python -I -m pip install --force-reinstall --no-deps /home/kurt/dev/pr-pypa-readme_renderer/.tox/.tmp/package/45/readme_renderer-43.0.tar.gz [tox/tox_env/api.py:425]
Processing ./.tox/.tmp/package/45/readme_renderer-43.0.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: readme_renderer
  Building wheel for readme_renderer (pyproject.toml) ... done
  Created wheel for readme_renderer: filename=readme_renderer-43.0-py3-none-any.whl size=13304 sha256=534ec0e0a66e38098ca98b4c2043c07f33557c0483c4604e8c1ad623694416c9
  Stored in directory: /home/kurt/.cache/pip/wheels/b3/1f/e0/093a05dd510eee76ab487c953f8603b49fe1ae082ce051dd60
Successfully built readme_renderer
Installing collected packages: readme_renderer
  Attempting uninstall: readme_renderer
    Found existing installation: readme_renderer 43.0
    Uninstalling readme_renderer-43.0:
      Successfully uninstalled readme_renderer-43.0
Successfully installed readme_renderer-43.0
py312: 3266 I exit 0 (2.39 seconds) /home/kurt/dev/pr-pypa-readme_renderer> python -I -m pip install --force-reinstall --no-deps /home/kurt/dev/pr-pypa-readme_renderer/.tox/.tmp/package/45/readme_renderer-43.0.tar.gz pid=103997 [tox/execute/api.py:286]
```

:point_up: 2.39 seconds to convert the `.tar.gz` file to a `.whl` and install it

### Package installation step (after)

```
py312: 861 W install_package> python -I -m pip install --force-reinstall --no-deps /home/kurt/dev/pr-pypa-readme_renderer/.tox/.tmp/package/46/readme_renderer-43.0-py3-none-any.whl [tox/tox_env/api.py:425]
[...]

PR #313, opened 2024-06-29
Skip installing the package when type linting

kurtmckee:
Mypy is configured to scan the source directory, not the installed package. Therefore, this PR configures tox to skip package installation.

On my machine, this change reduces the `mypy` tox run from 2.79 seconds to 0.25 seconds.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,6 +19,7 @@ extras = md
 
 [testenv:mypy]
 basepython = python3
+skip_install = true
 deps =
     mypy
     types-bleach


miketheman (approved):

kurtmckee:
_Rebased on `main`_

miketheman:
This appears to be failing as the runtime dependency `nh3` is not being installed, and it contains the type hints that `mypy` is looking for.
Is there another way to tell tox to install the package deps without installing the package itself?

kurtmckee:
Fascinating. Installing the package allows the dependencies to be available for mypy. On my machine, the cached `mypy` tox environment wasn't recreated when making this change, so the dependencies were still available and didn't trigger an error. Now that it's running in a clean environment in CI, the test is failing.

I'm closing this, as it's currently necessary to install the package to get the package dependencies that mypy wants to also have installed.

kurtmckee:
@miketheman We were typing at the same time. :smile: 

I don't know of a way to get the dependencies in a supported -- and non-duplicative -- way from within tox.

Tox supports `deps`, and nh3 _et al_ could be added there manually, but that adds a risk of administrative burden to maintain dependencies in `pyproject.toml` and `tox.ini`.

The environment setup times will be less costly if #308 merges.

PR #314, opened 2024-06-29
Lazy open output files, and always close them

kurtmckee:
This PR modifies the test suite to show all warnings, including `ResourceWarning`. By doing so, this revealed that the CLI code never closes output files opened by `argparse` during argument parsing.

Therefore, this PR additionally modifies the CLI code to open output files _only when they will be written to_, and ensures that they are always closed after writing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,7 @@ setenv =
     # Display up to 20 frames in backtraces when showing ResourceWarnings.
     PYTHONTRACEMALLOC=20
 commands =
-    python -Wall -m pytest --strict-markers --cov {posargs}
+    pytest -Wall --strict-markers --cov {posargs}
 extras = md
 
 [testenv:mypy]
--- a/tox.ini
+++ b/tox.ini
@@ -15,8 +15,11 @@ deps =
     pytest
     pytest-cov
     pytest-icdiff
+setenv =
+    # Display up to 20 frames in backtraces when showing ResourceWarnings.
+    PYTHONTRACEMALLOC=20
 commands =
-    pytest --strict-markers --cov {posargs}
+    python -Wall -m pytest --strict-markers --cov {posargs}
 extras = md
 
 [testenv:mypy]
--- a/readme_renderer/__main__.py
+++ b/readme_renderer/__main__.py
@@ -18,7 +18,7 @@ def main(cli_args: Optional[List[str]] = None) -> None:
                         help="README format (inferred from input file name or package)")
     parser.add_argument('input', help="Input README file or package name")
     parser.add_argument('-o', '--output', help="Output file (default: stdout)",
-                        type=argparse.FileType('w'), default='-')
+                        default='-')
     args = parser.parse_args(cli_args)
 
     content_format = args.format
@@ -55,7 +55,11 @@ def main(cli_args: Optional[List[str]] = None) -> None:
                          "`rst`, or `txt`)")
     if rendered is None:
         sys.exit(1)
-    print(rendered, file=args.output)
+    if args.output == "-":
+        print(rendered, file=sys.stdout)
+    else:
+        with open(args.output, "w") as fp:
+            print(rendered, file=fp)
 
 
 if __name__ == '__main__':


kurtmckee:
Rebased on `main`

tox.ini
```diff
@@ -15,8 +15,11 @@ deps =
     pytest
     pytest-cov
     pytest-icdiff
+setenv =
+    # Display up to 20 frames in backtraces when showing ResourceWarnings.
+    PYTHONTRACEMALLOC=20
 commands =
-    pytest --strict-markers --cov {posargs}
+    python -Wall -m pytest --strict-markers --cov {posargs}
```
miketheman, line 22: Curious why you went with the outer call instead of using `pytest -Wall ...`?
See https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html

miketheman (commented):
One question inline, but otherwise looks fine.

I'm curious about the impact of leaving open files in a CLI tool - don't those descriptors disappear once the runtime exits?

Considering the primary use cases for this project is a library, how important is this change?

kurtmckee:
Great questions.

* _I'm curious about the impact of leaving open files in a CLI tool - don't those descriptors disappear once the runtime exits?_

  Yes, they will. However, I had two considerations when I observed this.

  The first is a recent experience at work with a tool that expected to be run as a CLI but that we needed to import, customize, and call the `cli.main()` function ourselves. That recent experience made me want to address this as a hygiene/best practice consideration.

  The second is a functional consideration: **_the output file is created even if there's an error_**.

  For example, if nh3 raises `ValueError`, that is caught, `None` is returned, and `sys.exit(1)` is subsequently called. An output file is created even though rendering failed.

  This PR changes that behavior so an output file is only created when rendered output is ready to be written, and ensures that resources are cleaned up afterward.

* _how important is this change?_

  Not important. I think it's a low value -- but correct -- change.

* _Curious why you went with the outer call instead of using `pytest -Wall` ...?_

  I didn't know that `-Wall` existed! :grinning: 

  I have had past experience that configuring pytest's `filterwarnings` to raise errors on warnings wouldn't catch errors in very specific ways, like a `DeprecationWarning` that was triggered by [Python 3.12 when importing an old version of `dateutil`](https://github.com/dateutil/dateutil/issues/1284); the solution I found at that time was to switch from pytest's `filterwarnings` to invoking Python directly.

  Want me to switch to `pytest -Wall` instead of `python -Wall -m pytest`?

miketheman:
> ... we needed to import, customize, and call the `cli.main()` function ourselves ...

Thanks for the context, and that makes sense.

The library was never fully intended as a CLI, so please be mindful of the `cli` namespace, as it is entire possible to change (as we ware changing it here). I'd be open to understanding the utility performed in the CLI that should be part of the general API, and thus decrease usage on `cli.*`

> This PR changes that behavior so an output file is only created when rendered output is ready to be written, and ensures that resources are cleaned up afterward.

This makes most sense here.

>   Want me to switch to `pytest -Wall` instead of `python -Wall -m pytest`?

Yes please! If we have a specific reason to move it to the "outer" behavior you've described, then we can do that at a later time, but we should use the built-in if possible.

tox.ini
kurtmckee, line 22:
```suggestion
    pytest -Wall --strict-markers --cov {posargs}
```

miketheman (approved):

tox.ini
kurtmckee: I think this is addressed now.

kurtmckee:
Oh no, my commit signatures appear to have been broken by using GitHub's web UI to rebase on `main`. Let me circle back to this when I'm back at my PC tomorrow.

Regarding the `cli.main()` discussion -- I wanted to be clear that this was not something involving this project, it was just an example from my recent experience that "people use stuff in unexpected ways". 😀

miketheman:
Thanks! Merging now.

If you'd like to craft a CHANGES.rst entry for 44.0, we can get these released. Here's an example of a previous PR for releases: https://github.com/pypa/readme_renderer/pull/303/

PR #324, opened 2025-01-17
Support footnotes in Markdown

### LINKED ISSUE #323: Footnotes are not supported in Markdown
I used footnotes in [the project README](https://github.com/shenxianpeng/gitstats/tree/v0.4.1), which looks like the following.

Image

click [1] to jump to the footnotes

Image

---

Hower on https://pypi.org/project/gitstats, it will display like this which does not support footnotes 

Image

shenxianpeng:
closes #323

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -9,9 +9,9 @@ from unittest import mock
                         "test_GFM_001.md", "test_txt_001.txt"])
 def input_file(request):
     path = pathlib.Path("tests/fixtures", request.param)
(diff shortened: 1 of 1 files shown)


miketheman:
> From my test on the main branch, [the CI has failed already ](https://github.com/shenxianpeng/readme_renderer/actions/runs/12826998926/job/35768080327#step:5:289)without this change.

Yes, looks like a recent update to docutils or pygments may have changed the output - would you want to make a separate PR that only updates the test outputs based on current dependencies?

miketheman (approved):
Thanks!

PR #326, opened 2025-02-22
chore(dev): replace gitpod with devcontainer

miketheman:
As gitpod's moved to no longer cloud-based environments, and supports using the devcontainer standard, use that instead to support GitHub Codespaces.


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+Use the supplied [Development Container](https://containers.dev/) configuration
+to get a development environment up and running quickly.
+The configurations can be found in the `.devcontainer` directory.
+
+[GitHub Codespaces](https://github.com/features/codespaces) can provide
+a development environment in the cloud.
+GitHub Codespaces also provides a prebuilt image for this repository
+from the most recent commit on the `main` branch
+to speed up the setup process, including many of the Python runtimes
+and tools needed to run the tests.
```
miketheman, line 20: Note: This will only be true after it's been merged to `main` and enabled and Codespaces hasn't hit a spending limit.
https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/choosing-who-owns-and-pays-for-codespaces-in-your-organization

ewdurbin (approved):
seems, fine? do we have evidence of a need/desire for a devcontainer setup, or is this just another thing to maintain?

miketheman:
> seems, fine? do we have evidence of a need/desire for a devcontainer setup, or is this just another thing to maintain?

I used to use the gitpod approach for quicker, one-off fixes that wouldn't need me to have my local setup work with all the necessary Python versions, dependencies, etc. Once published and enabled with prebuild, could again be used in a similar fashion.

PR #327, opened 2025-07-24
Update the build-system requirements; resolve warnings and errors

kurtmckee:
# The current situation

The current build-system requires setuptools 40.8.0 or higher. However, setuptools v40.8.0 doesn't have a `build_meta` submodule.

To demonstrate the problem, the following commands were run:

```
python3.13 -m venv venv-3.13
source venv-3.13/bin/activate
python -m pip install setuptools==40.8.0 build
python -m build --no-isolation
```

This results in the following error message:

```
$ python -m build --no-isolation
* Getting build dependencies for sdist...

Traceback (most recent call last):
  File ".../lib/python3.13/site-packages/pyproject_hooks/_impl.py", line 402, in _call_hook
    raise BackendUnavailable(
    ...<4 lines>...
    )
pyproject_hooks._impl.BackendUnavailable: Cannot import 'setuptools.build_meta'

ERROR Backend 'setuptools.build_meta' is not available.
```

Upgrading to the latest version of setuptools, 80.9.0, allows the package to build:

```
python -m pip install --upgrade setuptools
python -m build --no-isolation
```

However, warnings are emitted:

```
SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
!!

********************************************************************************
Please use a simple string containing a SPDX expression for `project.license`.
You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

By 2026-Feb-18, you need to update your project and remove deprecated calls
or your builds will no longer be supported.

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
********************************************************************************
```

```
SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

********************************************************************************
Please consider removing the following classifiers in favor of a SPDX license expression:

License :: OSI Approved :: Apache Software License

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
********************************************************************************
```

# The solution

This change addresses the lower setuptools bound based on the warnings: it sets the setuptools minimum version to v77.0.1 (v77.0.0 was not released to PyPI).

In addition, it addresses the license-related warnings by switching to the SPDX short identifier `"Apache-2.0"` and by removing the `License` trove classifier.

With this change in place, the package can be built on the lowest supported setuptools version as well as the latest released version.

### DIFF
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -13,24 +13,24 @@
 # limitations under the License.
 
 import io
-from typing import Any, Dict, IO, Optional, Union
+from typing import Any, Dict, IO, Optional
 
 from docutils.core import publish_parts
-from docutils.nodes import colspec, image
+from docutils.nodes import Element
 from docutils.writers.html5_polyglot import HTMLTranslator, Writer
 from docutils.utils import SystemMessage
 
 from .clean import clean
 
 
-class ReadMeHTMLTranslator(HTMLTranslator):  # type: ignore[misc] # docutils is incomplete, returns `Any` python/typeshed#7256 # noqa E501
+class ReadMeHTMLTranslator(HTMLTranslator):
 
     # Overrides base class not to output `<object>` tag for SVG images.
     object_image_types: Dict[str, str] = {}
 
     def emptytag(
         self,
-        node: Union[colspec, image],
+        node: Element,
         tagname: str,
         suffix: str = "\n",
         **attributes: Any
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -44,13 +44,14 @@ def test_cli_invalid_format():
         main(["no-file.invalid"])
 
 
-def test_cli_explicit_format(input_file):
+def test_cli_explicit_format(input_file, tmp_path):
     fmt = input_file.suffix.lstrip(".")
-    with input_file.open() as fp, \
-            mock.patch("pathlib.Path.open", return_value=fp), \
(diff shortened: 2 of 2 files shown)


di (approved):
LGTM -- seems like there are unrelated linting failures on `main` right now, however.

pyproject.toml
```diff
@@ -48,7 +47,7 @@ exclude = ["tests", "tests.*"]
 Home-page = "https://github.com/pypa/readme_renderer"
 
 [build-system]
-requires = ["setuptools>=40.8.0"]
+requires = ["setuptools>=77.0.1"]
```
DimitriPapadopoulos, line 50: I would suggest `"setuptools>=77.0.3"` for consistency with the [Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files) or `"setuptools>=77.0.0"` for consistency with the [Setuptools documentation](https://setuptools.pypa.io/en/stable/userguide/pyproject_config.html).

I understand that in theory 77.0.0 supports PEP 639, but in practice 77.0.3 fixes known early bugs of the PEP 639 implementation. But then 77.0.1 might be good enough for this specific use case.

kurtmckee: I've updated to require `setuptools>=77.0.3`. Thanks for suggesting an update that resolves some known bugs!

kurtmckee:
@di If you didn't see it, I posted PR #332 to address the failing test case.

PR #330, opened 2025-08-05
Apply ruff/pyupgrade rules (UP)

DimitriPapadopoulos:
I would suggest testing with ruff in addition to, or instead of flake8. This prepares such a change.

~~Requires #328 and #332 to pass CI tests.~~

### DIFF
--- a/readme_renderer/__main__.py
+++ b/readme_renderer/__main__.py
@@ -5,10 +5,10 @@ from readme_renderer.txt import render as render_txt
 import pathlib
 from importlib.metadata import metadata
 import sys
-from typing import Optional, List
+from typing import Optional
 
 
-def main(cli_args: Optional[List[str]] = None) -> None:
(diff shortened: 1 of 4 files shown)


miketheman (changes requested):
A lot of these changes are not compliant with the minimum supported version of Python

DimitriPapadopoulos:
> A lot of these changes are not compliant with the minimum supported version of Python

That would be surprising, Ruff automatically takes into account the minimum supported version of Python:
https://github.com/pypa/readme_renderer/blob/d184e41dc9c19ba18aa3c1fc72025fe2ce19a4e1/pyproject.toml#L36

The CI errors might have been caused by a human error. Let me have a look.

DimitriPapadopoulos:
Indeed a rebase error. Fixed now.

miketheman (approved):
Thanks!

PR #331, opened 2025-08-08
PEP 639 compliance

DimitriPapadopoulos:


kurtmckee:
This is a duplicate of #327.

PR #332, opened 2025-08-20
Fix a test failure caused by an internal docutils `open()` call

kurtmckee:
docutils 0.22 introduced an RST stylesheet feature that must be able to open files during RST-to-HTML rendering.

The original design of the `test_cli_explicit_format` test patched `pathlib.Path.open()` so it would always return the same open file instance...which was always closed on first use. This caused a failure when docutils tried to open `minimal.css`:

```
ValueError: I/O operation on closed file.
```

This change introduces a more complex mocking strategy that documents and meets the current technical requirements.

When combined with #328, the full test suite will now pass.

> [!NOTE]
>
> It may be worthwhile to consider using the pyfakefs plugin, and pytest's capsys, to simplify some of the mocking and patching used in the test suite.

### DIFF
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -13,24 +13,24 @@
 # limitations under the License.
 
 import io
-from typing import Any, Dict, IO, Optional, Union
+from typing import Any, Dict, IO, Optional
 
 from docutils.core import publish_parts
-from docutils.nodes import colspec, image
+from docutils.nodes import Element
 from docutils.writers.html5_polyglot import HTMLTranslator, Writer
 from docutils.utils import SystemMessage
 
 from .clean import clean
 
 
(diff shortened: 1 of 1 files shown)


kurtmckee:
Hello @miketheman, I hope you're well! I was wondering if you could take a look at this PR -- and a couple of others -- that address lint and test suite failures?

For example, this PR fixes new CI failures; #328 addresses lint failures; and both of these issues were found when CI ran on #327.

Thanks for your time, and have a great weekend!

kurtmckee:
Works for me! I had been under the impression that it was undesirable to write a temporary file so I tried to stick with that assumed goal. Much easier this way!

I've updated the PR to match your suggestion. Type annotation linting will fail; that's addressed in #328.

PR #334, opened 2025-10-10
Test PyPy 3.11

DimitriPapadopoulos:
[PyPy](https://pypy.org/) 3.11 is the currently supported version:
> We currently support python 3.11 and 2.7.

Perhaps we should get rid of PyPy 3.9 and 3.10, but let's defer to a later PR or commit.


miketheman:
Thanks for digging those up!

PR #336, opened 2025-10-10
Apply ruff (RUF) and tryceratops (TRY) rules

DimitriPapadopoulos:
I would suggest testing with ruff in addition to, or instead of flake8. This prepares such a change.

### DIFF
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -83,7 +83,6 @@ def clean(
             link_rel="nofollow",
             url_schemes={"http", "https", "mailto"},
         )
-
-        return cleaned
     except ValueError:
         return None
+    return cleaned
--- a/readme_renderer/rst.py
+++ b/readme_renderer/rst.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import io
-from typing import Any, IO, Optional
+from typing import Any, ClassVar, IO, Optional
 
 from docutils.core import publish_parts
 from docutils.nodes import Element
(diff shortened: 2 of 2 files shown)


zercao012-source (approved):

readme_renderer/clean.py
```diff
@@ -83,7 +83,7 @@ def clean(
             link_rel="nofollow",
             url_schemes={"http", "https", "mailto"},
         )
-
-        return cleaned
     except ValueError:
         return None
+    else:
+        return cleaned
```
miketheman: I think that if we're going to move it, let's remove the `else` return cleaned as the final operation, unless there's a reason to keep it?

readme_renderer/rst.py
```diff
@@ -26,7 +26,7 @@
 class ReadMeHTMLTranslator(HTMLTranslator):
 
     # Overrides base class not to output `<object>` tag for SVG images.
-    object_image_types: Dict[str, str] = {}
+    object_image_types: ClassVar[Dict[str, str]] = {}
```
miketheman: I think there's a conflict here, rebase needed.

miketheman (changes requested):
Generally fine.

readme_renderer/clean.py
DimitriPapadopoulos: Done.

readme_renderer/rst.py
DimitriPapadopoulos: Rebased.

miketheman (approved):
Thanks!

PR #337, opened 2025-10-15
Migrate license configuration from setup.cfg to pyproject.toml

shenxianpeng:
This PR removes the legacy `setup.cfg` file and moves license configuration to `pyproject.toml`, using the modern [PEP 621](https://peps.python.org/pep-0621/) – compliant metadata field [license-files](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-files).


miketheman (approved):
Thanks!

PR #351, opened 2026-01-22
feat: support inline link anchors

### LINKED ISSUE #71: Should warn about relative URLs in links and images
Such relative URLs will not work correctly when displayed on PyPI.

### LINKED ISSUE #169: Support internal links in markdown long_descriptions
**What's the problem this feature will solve?**
I want to set internal links/anchors in my README.md/long_description in markdown format (for example `See [the documentation](#documentation) for more information` which links to a `## Documentation` heading or an explicit `<a name="documentation"></a>` anchor). Ideally I'd like to use the same internally linked Markdown source for both GitHub and PyPI.

**Describe the solution you'd like**
Explicit link creation via `<a name>` and `<a id>` should be supported (maybe only `<a name>`, depending on the potential to disturb any Javascript code).

Ideally GitHub-compatible automatic anchor generation for headings should also be supported.

**Additional context**
A GitHub README.md supports two ways of setting anchors:
* Implicitly, by generating an anchor from a heading (an `<a id="lower-case-with-some-chars-replaced-by-hyphens">`)
* Explicitly, by placing `<a id="...">` or `<a name="..."`> tags.

Warehouse does not automatically generate anchors for headings. It also strips `id` and `name` attributes from HTML embedded in the Markdown file. This leaves no possibility of linking to another location within the description (without using an absolute link to PyPI.org, which is undesirable since the README is typically used in other places like GitHub too).

miketheman:
Refs: https://docs.rs/comrak/latest/comrak/options/struct.Extension.html#structfield.header_ids
Refs: https://docutils.sourceforge.io/docs/user/config.html#section-self-link

### DIFF
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -35,6 +35,7 @@ try:
     gfm_extension_options = comrak.ExtensionOptions()
     gfm_extension_options.autolink = True
     gfm_extension_options.footnotes = True
+    gfm_extension_options.header_ids = ""  # intentionally blank
     gfm_extension_options.strikethrough = True
     gfm_extension_options.table = True
     gfm_extension_options.tagfilter = True
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -29,13 +29,17 @@ _EXTRA_WARNING = (
     "Install 'readme_renderer[md]' to enable Markdown rendering."
 )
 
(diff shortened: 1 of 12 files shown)


miketheman:
@di I believe I've addressed the approach to add a `user-content-` prefix in both rst and md renderings.

di (approved):

PR #355, opened 2026-04-10
Run ruff in addition to flake8 during CI

DimitriPapadopoulos:


tox.ini
```diff
@@ -3,6 +3,7 @@ envlist =
     py{310, 311, 312, 313, 314}
     pypy{311}
     pep8
+    ruff
```
DimitriPapadopoulos, line 6: Is that the proper way to enable ruff runs?

miketheman: Almost - I think one more place is needed:
https://github.com/pypa/readme_renderer/blob/a2c38618a9db23eac5eee06dd5bfafd4aa5606c0/.github/workflows/ci.yml#L56-L59

DimitriPapadopoulos: Tried this:
```yaml
 - name: Run linting 
   run: python -m tox -e pep8 -e ruff
```

miketheman (approved):

PR #356, opened 2026-04-10
Use well-known labels

DimitriPapadopoulos:
See:
* [PEP 753 – Uniform project URLs in core metadata](https://peps.python.org/pep-0753/)
* [Well-known Project URLs in Metadata](https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels)
* [Home-page](https://packaging.python.org/en/latest/specifications/core-metadata/#home-page)


miketheman (approved):
Thanks!

PR #361, opened 2026-04-28
Fix CLI --package flag for Python 3.15+

hrnciar:
Python 3.15 synced importlib.metadata with importlib_metadata 8.7, which stores the package description in the Description header field instead of the message payload. Try the Description header first, falling back to get_payload() for older Python versions.

Assisted-by: Cursor

### DIFF
--- a/readme_renderer/__main__.py
+++ b/readme_renderer/__main__.py
@@ -24,7 +24,7 @@ def main(cli_args: Optional[list[str]] = None) -> None:
     content_format = args.format
     if args.package:
         message = metadata(args.input)
-        source = message.get_payload()  # type: ignore[attr-defined] # noqa: E501 https://peps.python.org/pep-0566/
+        source = message.get("Description") or message.get_payload()  # type: ignore[attr-defined] # noqa: E501 https://peps.python.org/pep-0566/
 
         # Infer the format of the description from package metadata.
         if not content_format:


miketheman (commented):
Thanks - I don't know that this is ready to be tested with Python 3.15 yet - Maybe add a commit that expands testing for 3.15 first, so we see the failure, and then this change proves it fixes it.

readme_renderer/__main__.py
```diff
@@ -24,7 +24,7 @@ def main(cli_args: Optional[list[str]] = None) -> None:
     content_format = args.format
     if args.package:
         message = metadata(args.input)
-        source = message.get_payload()  # type: ignore[attr-defined] # noqa: E501 https://peps.python.org/pep-0566/
+        source = message.get("Description") or message.get_payload()  # type: ignore[attr-defined] # noqa: E501 https://peps.python.org/pep-0566/
```
hroncok, line 27: FTR, this seems to work on all supported Python versions:

```suggestion
        source = message.get("Description")
```

hrnciar:
> Thanks - I don't know that this is ready to be tested with Python 3.15 yet - Maybe add a commit that expands testing for 3.15 first, so we see the failure, and then this change proves it fixes it.

I've opened https://github.com/pypa/readme_renderer/pull/375, I suppose the CI will fail with 3.15.

PR #366, opened 2026-05-27
Restrict OIDC token to publish job

facutuesca:
The GH OIDC token used for Trusted Publishing should not be available to non-publishing steps. This PR removes it from the steps that install the dependencies and build the project, so that it's only available during PyPI publishing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,11 +6,9 @@ on:
 permissions: {}
 
 jobs:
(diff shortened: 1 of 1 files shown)


miketheman:
Is this a duplicate of https://github.com/pypa/readme_renderer/pull/363 ?

facutuesca:
ah yes, sorry I missed it. Closing this one now

PR #367, opened 2026-06-02
Strip non-task-list input controls

facutuesca:
Currently, active input controls are accepted and rendered:

image

--------------

This is a regression introduced in the [migration](https://github.com/pypa/readme_renderer/pull/295) from Bleach to nh3 (see the removed `DisabledCheckboxInputsFilter` class)

This PR fixes it by adding an `HTMLParser` to get the location of these inputs and a function to remove them when calling `clean()`.

cc @miketheman

### DIFF
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


facutuesca:
Superseded by : https://github.com/pypa/readme_renderer/pull/370

PR #368, opened 2026-06-02
Fix ignored allowlists on `clean()`

facutuesca:
Custom tags and attributes passed to `clean()` were ignored and always replaced by the default values. This PR fixes that

cc @miketheman


sethmlarson (approved):
LGTM! Thank you :)

miketheman (approved):
Thanks!

PR #370, opened 2026-06-04
fix: disable input controls

miketheman:
The Bleach-to-nh3 migration dropped the filter that stripped active <input> controls, so submit buttons, text fields, and enabled checkboxes came back. ~~Now nh3 forces disabled on every input and drops any type that isn't checkbox, so only the task-list checkboxes survive - and they're inert.~~ Avoids re-parsing the already-sanitized HTML.
Adds a `disabled` attribute to all input controls.


miketheman:
Simpler alternative to #367

facutuesca:
> Simpler alternative to #367



This doesn't have the same behavior as #367. For example:
```
<input type="radio" checked>
```

becomes:
```
<input disabled="" checked="">
```

which shows a disabled text box
image

miketheman:
~~That is probably because the 'radio' attribute isn't allowed through.
What does the main branch do with 'radio' today?~~ My filter is too specific, I'll look at expanding it to all input types.

miketheman:
While it's rare that anyone uses radio boxes or other inputs in their readme files as there's no direct markdown control for that, both rst and markdown may contain raw HTML and can express `input` directives.

Because `<input type=...` is allowed, disable all possible inputs, as the rendered readme is not meant to be an interactive HTML page.

facutuesca:
> While it's rare that anyone uses radio boxes or other inputs in their readme files as there's no direct markdown control for that, both rst and markdown may contain raw HTML and can express `input` directives.
> 
> Because `<input type=...` is allowed, disable all possible inputs, as the rendered readme is not meant to be an interactive HTML page.

Yeah, this LGTM!

ewdurbin (approved):

PR #374, opened 2026-06-14
feat: enable GFM alerts and shortcodes

### LINKED ISSUE #318: PyPI does not render GFM alerts
The alerts that display on the main page of the repository are not displaying properly on PyPI

GitHub: https://github.com/tektronix/tm_devices
PyPI: https://pypi.org/project/tm-devices/

miketheman:
Two feature enhancements, split by commit, enabled by extension toggle.

---
feat: enable GFM-style alerts/admonitions 

Instead of the `blockquote` that it is rendering now buy virtue of the
leading `>`, activate the extension and set the correct classes.

Will still need styling on target sites, but now there's classes to
leverage.

Resolves #318

---
feat: support shortcodes/emoji

Supports whatever upstream can provide.

Refs: https://github.com/rossmacarthur/emojis
Refs: https://github.com/github/gemoji


ewdurbin (approved):

shenxianpeng (approved):

flying-sheep:
> Will still need styling on target sites, but now there's classes to
leverage.

We already had classes through rST admonitions.

The code added in this PR should be changed to use them before adding styling, see #378

PR #379, opened 2026-07-11
fix: unify admonition HTML

### LINKED ISSUE #378: Use existing admonition tags and classes in new GFM admonitions
In #374 you added admoniton support but didn’t match the existing rST HTML structure that exists in many READMEs on PyPI already.

We should change that and emit the existing structure instead so existing READMEs can benefit from the new to-be-added styling.

e.g. for [django-push-notifications](https://pypi.org/project/django-push-notifications/), this

```rst
.. note::
	To migrate from legacy FCM APIs to HTTP v1, see `docs/FCM <https://github.com/jazzband/django-push-notifications/blob/master/docs/FCM.rst>`_.
```

becomes this HTML

```html
<aside class="admonition note">
<p class="admonition-title">Note</p>
<p>To migrate from legacy FCM APIs to HTTP v1, see <a href="https://github.com/jazzband/django-push-notifications/blob/master/docs/FCM.rst" rel="nofollow">docs/FCM</a>.</p>
</aside>
```

and renders as

image

the classes in this PR should be adjusted to match the rST classes that already exist before adding support.

cc @miketheman @ewdurbin

flying-sheep:
fixes #378

should be quite straightforward and robust:
- If tag names and classes match the two known tags emitted by comrak, we replace start and end tags with the desired HTML
- When handling the end tag, we just skip void (self-closing) tags that don’t match until we find a match. This makes it robust against HTML5 adding new void tags
- `self.get_starttag_text()` is used for unhandled tags, which means we can’t make mistakes re-emitting them

The only thing I’m unsure about is that we don’t handle declarations (`<!`) and processing instructions (`<?`): should we do that?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ classifiers = [
 requires-python = ">=3.10"
 
 [project.optional-dependencies]
-md = ["comrak>=0.0.11"]
+md = ["comrak>=0.0.13"]
 
 [tool.setuptools]
 include-package-data = true
--- a/readme_renderer/__main__.py
+++ b/readme_renderer/__main__.py
@@ -1,10 +1,11 @@
 import argparse
+import pathlib
+import sys
+from importlib.metadata import metadata
+
 from readme_renderer.markdown import render as render_md
 from readme_renderer.rst import render as render_rst
 from readme_renderer.txt import render as render_txt
-import pathlib
-from importlib.metadata import metadata
-import sys
 
 
 def main(cli_args: list[str] | None = None) -> None:
--- a/readme_renderer/clean.py
+++ b/readme_renderer/clean.py
@@ -14,7 +14,6 @@
 
 import nh3
 
-
 ALLOWED_TAGS = {
     # Bleach Defaults
     "a", "abbr", "acronym", "b", "blockquote", "code", "em", "i", "li", "ol",
--- a/readme_renderer/markdown.py
+++ b/readme_renderer/markdown.py
@@ -19,6 +19,7 @@ from collections.abc import Callable
 from re import Match
 
 from html import unescape
+from html.parser import HTMLParser
 
 import pygments
 import pygments.lexers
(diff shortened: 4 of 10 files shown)


miketheman:
Thanks for tackling this — unifying alert markup with the docutils output is clearly the right goal, and #822 looks like the clean way to get there.

My worry with this PR is where it sits in the pipeline. It parses and re-serializes the whole rendered document before `clean()` runs, and since comrak renders with `unsafe_=True`, `_GFMAlertRewriter` ends up handling arbitrary untrusted raw HTML from every README. A few things it does there concern me for code PyPI runs on every package description.

The one that actually worries me is a DoS. `handle_endtag` copies and scans the entire open-tag stack for every end tag, and void or unclosed tags never leave the stack, so it's O(n²). I measured a ~1 MB README of `<span>×N` followed by `</b>×N` — well under the `long_description` limit — at roughly 16 minutes of CPU. That's enough to pin Warehouse render workers.

There's also quiet content loss. The parser doesn't handle declarations, processing instructions, or CDATA, so `<?php … ?>` and `<![CDATA[…]]>` in a raw code block just disappear, even though they render fine on `main`. And the entity handlers append a semicolon that wasn't in the source, which can quietly turn `a &verbar b` in a code sample into `a | b`.

Last one: the alert type gets interpolated into the `<aside>` tag unescaped, so a crafted class value can break out and inject attributes, and a forged `id` makes it through `clean()`. Not stored XSS today, since nh3 runs last, but the rewriter is building attacker-shaped markup and trusting nh3 to catch everything, which I'd rather not rely on.

All of this comes from the same place: we're round-tripping hostile HTML through a hand-written serializer just to rename two classes. That's why I'd rather wait for comrak #822 and get the semantic HTML straight from the parser — no re-serialization pass, no serializer to maintain, and none of these edge cases. If we want something before then, at least running the rewrite after `clean()` (nh3 already allows `class` on `div`/`p`/`aside`) keeps the input well-formed and sanitized. And then we might have to run `clean()` a second time, to be certain nothing else crept in.

flying-sheep:
> That's why I'd rather wait for comrak #822 and get the semantic HTML straight from the parser 

yeah, already released! Now we just need to wait for lmmx/comrak#76 to get merged and released.

If that happens before we want to release all is good!

flying-sheep:
@miketheman here we go, updated

miketheman (approved):
Nicely done, well coordinated across multiple projects. Thanks for pushing this forward!

miketheman:
@flying-sheep unclear why we had lint fail post-merge - can you please examine? https://github.com/pypa/readme_renderer/actions/runs/30285070109/job/90040719817
otherwise I'll only have time to check in on this in a couple of weeks.

flying-sheep:
done. it’s

> error: Redundant cast to "str"

which is the best kind of error since it means “mypy used to be too limited to understand this code, now it’s good enough to figure it out”

you fix the error by removing the now-redundant casts.

